PR #37, opened 2026-03-07
feat(agent): add in-kernel NFSv4 server support

PeronGH:
Add end-to-end NFS service support: protobuf messages, wire types,
guest agent module, and host-side autostart with port forwarding.

The guest agent manages the in-kernel NFSv4 server via procfs,
exporting the Btrfs data drive as the v4 pseudo-root. The host
spawns a background task during runtime init to ensure NFS is
ready and forwards TCP port 2049.

- Add NfsEnsureRequest/Response proto messages and wire types (0x0005/0x1005)
- Add NFS_PORT constant (2049)
- Add nfs.rs guest module with idempotent start/probe via RuntimeGuard
- Wire EnsureNfs into RPC dispatcher and runtime status checks
- Add ensure_nfs client method and background autostart in runtime init

### DIFF
--- a/app/arcbox-core/src/agent_client.rs
+++ b/app/arcbox-core/src/agent_client.rs
@@ -8,8 +8,8 @@ use arcbox_constants::wire::{
     ERROR_HEADER_SIZE, FRAME_HEADER_SIZE, MessageType, TRACE_LEN_FIELD_SIZE, TYPE_FIELD_SIZE,
 };
 use arcbox_protocol::agent::{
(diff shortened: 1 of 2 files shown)


PeronGH:
Closing — superseded by the vsock-proxy approach on `feat/nfs-docker-export`.

The original design (direct NFS port forwarding via bridge NIC) doesn't work reliably:
- vmnet.framework times out on some machines
- VZNATNetworkDeviceAttachment fallback doesn't support inbound TCP
- `select_routable_ip` picks the NAT IP instead of the bridge IP

The new approach routes NFS through vsock (localhost TCP → vsock → guest relay → nfsd), bypassing bridge networking entirely. See ABX-155 for tracking.

PR #54, opened 2026-03-11
feat(sandbox): implement interactive exec with bidirectional I/O

lucas77778:
## Summary

Implements `arcbox sandbox exec` — a full bidirectional interactive
exec channel from CLI to sandbox process, across all layers of the stack.

## Changes

**Wire protocol** (`arcbox-constants`)
- Add `SandboxExecInput = 0x0033` — streaming stdin frame type; empty payload = EOF

**Transport** (`arcbox-transport`)
- Add `VsockSender` / `VsockReceiver` half-types backed by independent fds via `libc::dup`
- Add `VsockTransport::into_split()` — enables concurrent reads and writes from separate tasks without a shared mutex

**Guest agent** (`arcbox-agent`)
- Add `SandboxService::exec()` — pre-encodes `ExecOutput` to proto bytes, same pattern as `run()`
- Add `handle_sandbox_exec()` — `tokio::select!` loop bridging vm-agent output and host stdin frames

**Core** (`arcbox-core`)
- Add `AgentClient::sandbox_exec()` — sends `SandboxExecRequest`, splits transport, spawns stdin pump + output pump tasks

**API** (`arcbox-api`)
- Implement gRPC `exec()` endpoint — reads `Init` from first message, forwards `Stdin`/`Resize` frames, streams `ExecOutput` back

**CLI** (`arcbox-cli`)
- Add `arcbox sandbox exec <id> <cmd...>` with `-t`/`--tty` and `--timeout` flags
- TTY mode: raw terminal, SIGWINCH → `Resize` frames, exit code propagation

## Test plan

- [ ] `arcbox sandbox exec <id> -- echo hello` outputs and exits cleanly
- [ ] `arcbox sandbox exec -t <id> -- /bin/sh` opens an interactive shell
- [ ] Ctrl+C / EOF closes the session and propagates exit code
- [ ] Terminal resize (SIGWINCH) is forwarded correctly in TTY mode

### DIFF
--- a/virt/arcbox-vm/src/bin/vm-agent.rs
+++ b/virt/arcbox-vm/src/bin/vm-agent.rs
@@ -651,7 +651,47 @@ mod agent {
     // vsock listener
     // -------------------------------------------------------------------------
 
+    /// Mount essential virtual filesystems for an init process.
+    fn mount_filesystems() {
+        use std::ffi::CString;
+
+        let mounts: &[(&str, &str, &str, libc::c_ulong, &str)] = &[
+            ("/proc", "proc", "proc", libc::MS_NOSUID | libc::MS_NODEV | libc::MS_NOEXEC, ""),
+            ("/sys", "sysfs", "sysfs", libc::MS_NOSUID | libc::MS_NODEV | libc::MS_NOEXEC, ""),
+            ("/dev", "devtmpfs", "devtmpfs", libc::MS_NOSUID, "mode=0755"),
+            ("/dev/pts", "devpts", "devpts", libc::MS_NOSUID | libc::MS_NOEXEC, "newinstance,ptmxmode=0666"),
+        ];
+
+        for (target, source, fstype, flags, data) in mounts {
+            let _ = std::fs::create_dir_all(target);
+            let c_source = CString::new(*source).unwrap();
+            let c_target = CString::new(*target).unwrap();
+            let c_fstype = CString::new(*fstype).unwrap();
+            let c_data = CString::new(*data).unwrap();
+            // SAFETY: all pointers are valid C strings.
+            let ret = unsafe {
+                libc::mount(
+                    c_source.as_ptr(),
+                    c_target.as_ptr(),
+                    c_fstype.as_ptr(),
+                    *flags,
+                    c_data.as_ptr().cast(),
+                )
+            };
+            if ret != 0 {
+                eprintln!(
+                    "mount {target}: {}",
+                    std::io::Error::last_os_error()
+                );
+            }
+        }
+
+        // Symlink /dev/ptmx → /dev/pts/ptmx so openpty() works.
+        let _ = std::os::unix::fs::symlink("/dev/pts/ptmx", "/dev/ptmx");
+    }
+
     pub fn run() {
+        mount_filesystems();
         eprintln!(
             "vmm-guest-agent: listening on vsock ports {AGENT_PORT} (exec), {FILE_PORT} (file I/O)"
         );


hypervisor/arcbox-vm/src/bin/vm-agent.rs
```diff
@@ -182,29 +194,185 @@ mod agent {
...
+        let secs = i64::from_le_bytes(payload[..8].try_into().unwrap());
+        let nanos = u32::from_le_bytes(payload[8..12].try_into().unwrap());
+
+        // SAFETY: clock_settime requires CAP_SYS_TIME; vm-agent runs as root inside guest.
+        unsafe {
+            let ts = libc::timespec {
+                tv_sec: secs,
+                tv_nsec: nanos as libc::c_long,
+            };
+            libc::clock_settime(libc::CLOCK_REALTIME, &raw const ts);
```
AprilNEA: `clock_settime` return value is ignored — the agent always responds with `MSG_EXIT(0)` even if the syscall fails (e.g., missing `CAP_SYS_TIME`). Check the return value and send `MSG_EXIT(-1)` on failure.

```suggestion
            let ret = libc::clock_settime(libc::CLOCK_REALTIME, &raw const ts);
            if ret != 0 {
                eprintln!(
                    "agent: clock_settime failed: {}",
                    std::io::Error::last_os_error()
                );
                let _ = write_frame(&mut conn, MSG_EXIT, &(-1i32).to_le_bytes());
                return;
            }
```

hypervisor/arcbox-vm/src/vsock.rs
```diff
@@ -360,6 +373,56 @@ pub async fn exec(
...
+    let nanos = now.subsec_nanos();
+    let mut payload = [0u8; 12];
+    payload[..8].copy_from_slice(&secs.to_le_bytes());
+    payload[8..].copy_from_slice(&nanos.to_le_bytes());
+
+    write_frame(&mut stream, MSG_CLOCK_SYNC, &payload)
+        .await
+        .map_err(|e| VmmError::Vsock(format!("write MSG_CLOCK_SYNC: {e}")))?;
+
+    let (msg_type, payload) = read_frame(&mut stream)
```
AprilNEA: `sync_clock()` awaits a response frame with no timeout. If the guest agent accepts the connection but never replies, the restore path hangs at this `await` indefinitely. Wrap in `tokio::time::timeout`.

```suggestion
    let (msg_type, payload) = tokio::time::timeout(
        Duration::from_secs(5),
        read_frame(&mut stream),
    )
    .await
    .map_err(|_| VmmError::Vsock("clock sync: timed out waiting for response".into()))?
    .map_err(|e| VmmError::Vsock(format!("read clock sync response: {e}")))?;
```

app/arcbox-core/src/agent_client.rs
```diff
@@ -578,6 +578,113 @@ impl AgentClient {
...
+            .await
+            .map_err(|e| CoreError::Machine(format!("failed to send exec request: {}", e)))?;
+
+        let (mut sender, mut receiver) = self
+            .transport
+            .into_split()
+            .map_err(|e| CoreError::Machine(format!("failed to split transport: {}", e)))?;
+
+        let (in_tx, mut in_rx) = mpsc::channel::<Vec<u8>>(16);
+        let (out_tx, out_rx) = mpsc::unbounded_channel();
```
AprilNEA, line 617: `mpsc::unbounded_channel()` for output means if the consumer is slow (e.g., CLI writing to a slow terminal), backpressure is not applied and memory grows unboundedly. For large command outputs (e.g., `cat /dev/urandom`), this could OOM the host. Use a bounded channel instead.

hypervisor/arcbox-vm/src/bin/vm-agent.rs
```diff
@@ -182,29 +194,185 @@ mod agent {
...
+                return;
+            }
+        };
+        let mode = if req.mode == 0 { 0o644 } else { req.mode };
+
+        // Collect FILE_DATA chunks until FILE_DONE.
+        let mut data: Vec<u8> = Vec::new();
+        loop {
+            match read_frame(&mut conn) {
+                Ok((FILE_DATA, chunk)) => data.extend_from_slice(&chunk),
```
AprilNEA: The file write handler accumulates data in memory with no total size limit. A buggy or malicious host could stream unlimited data and OOM the guest. Add a `MAX_FILE_SIZE` constant (e.g., 256 MiB) and check `data.len()` after each `extend_from_slice`, returning `FILE_ERR` if exceeded.

hypervisor/arcbox-vm/src/file_io.rs
```diff
@@ -0,0 +1,163 @@
...
+        .map_err(|e| VmmError::Vsock(format!("write FILE_READ_REQ: {e}")))?;
+
+    // Collect FILE_DATA chunks until FILE_DONE or FILE_ERR.
+    let mut buf = Vec::new();
+    loop {
+        let (frame_type, payload) = read_frame(&mut stream)
+            .await
+            .map_err(|e| VmmError::Vsock(format!("read file data: {e}")))?;
+        match frame_type {
+            FILE_DATA => buf.extend_from_slice(&payload),
```
AprilNEA: The file I/O protocol implementation accumulates the entire file in memory on the host (`buf.extend_from_slice` in a loop) with no size limit. Consider enforcing a maximum file size (aligned with the agent) or exposing a streaming API (e.g., `AsyncRead`) so large files can't OOM the host process.

guest/arcbox-agent/src/sandbox.rs
```diff
@@ -166,7 +170,151 @@ impl SandboxService {
...
+    /// - Read pre-encoded [`sandbox_v1::ExecOutput`] payloads from
+    ///   `output_receiver`.  The final payload has `done == true`.
+    pub async fn exec(
+        &self,
+        payload: &[u8],
+    ) -> Result<(mpsc::Sender<ExecInputMsg>, mpsc::UnboundedReceiver<Vec<u8>>), String> {
+        let req = sandbox_v1::ExecRequest::decode(payload)
+            .map_err(|e| format!("decode error: {e}"))?;
+
+        let tty_size = req.tty_size.map(|s| (s.width as u16, s.height as u16));
```
AprilNEA, line 214: `s.width as u16` silently truncates if the protobuf `u32` value exceeds `u16::MAX`. Use `u16::try_from()` with a fallback to avoid surprising behavior.

```suggestion
        let tty_size = req.tty_size.map(|s| {
            (
                u16::try_from(s.width).unwrap_or(80),
                u16::try_from(s.height).unwrap_or(24),
            )
        });
```

app/arcbox-cli/src/commands/sandbox.rs
```diff
@@ -437,6 +459,163 @@ async fn execute_run(args: RunArgs) -> Result<()> {
...
+                timeout_seconds: args.timeout,
+                ..Default::default()
+            })),
+        })
+        .await
+        .context("Failed to send exec init")?;
+
+    // Enable raw terminal mode when TTY is requested.
+    let _raw_guard = if args.tty {
+        Some(RawModeGuard::new()?)
```
AprilNEA, line 496: Raw mode is enabled before the gRPC `exec()` call at line 580. If the gRPC call fails, the error message from `context("Failed to exec in sandbox")` is printed while the terminal is in raw mode, producing garbled output. Move the `RawModeGuard` activation to after the gRPC call succeeds.

app/arcbox-api/src/grpc.rs
```diff
@@ -354,11 +354,65 @@ impl SandboxService for SandboxServiceImpl {
...
+            while let Some(msg) = stream.next().await {
+                match msg {
+                    Ok(input) => match input.payload {
+                        Some(exec_input::Payload::Stdin(data)) => {
+                            if in_tx.send(data).await.is_err() {
+                                return;
+                            }
+                        }
+                        Some(exec_input::Payload::Resize(_)) => {
+                            // TTY resize not yet forwarded over the wire protocol.
```
AprilNEA, line 396: TTY resize is silently dropped. The CLI already sends `Resize` frames in TTY mode, so users will observe that terminal resizes have no effect with no indication of why. At minimum, add a `warn!` log here so the gap is diagnosable. Better yet, forward the resize to the agent since the vm-agent already supports `MSG_RESIZE`.

guest/arcbox-agent/src/sandbox.rs
```diff
@@ -166,7 +170,151 @@ impl SandboxService {
...
+        loop {
+            tokio::select! {
+                // Output from vm-agent → host.
+                encoded_opt = out_rx.recv() => {
+                    match encoded_opt {
+                        None => break, // output channel closed
+                        Some(encoded) => {
+                            // The last frame has done==true; check before writing
+                            // so we can break after sending it.
+                            let done = sandbox_v1::ExecOutput::decode(encoded.as_slice())
```
AprilNEA, line 301: The output bytes were already encoded from `ExecOutput` at line 228, and here they are re-decoded just to check `m.done`. This is wasteful. Consider either:
1. Returning `(Vec<u8>, bool)` from the encoding task so `done` is passed alongside the pre-encoded bytes, or
2. Making the output channel carry structured `ExecOutput` and encoding at the write site.

app/arcbox-api/src/grpc.rs
```diff
@@ -354,11 +354,65 @@ impl SandboxService for SandboxServiceImpl {
...
+        let machine = extract_machine_id(&request)?;
+        let agent = self
+            .runtime
+            .get_agent(&machine)
+            .map_err(|e| core_to_status(&e))?;
+
+        let mut stream = request.into_inner();
+
+        // The first message in the stream must carry the Init payload.
+        let first = stream
```
AprilNEA, line 368: `stream.next().await` blocks indefinitely waiting for the first `Init` message. A misbehaving client that opens the stream but never sends data will hold this handler forever. Consider wrapping in `tokio::time::timeout`.

hypervisor/arcbox-vm/src/vsock.rs
```diff
@@ -360,6 +373,56 @@ pub async fn exec(
...
+        .await
+        .map_err(|e| VmmError::Vsock(format!("read clock sync response: {e}")))?;
+
+    if msg_type != MSG_EXIT {
+        return Err(VmmError::Vsock(format!(
+            "clock sync: unexpected response type 0x{msg_type:02x}"
+        )));
+    }
+    let code = if payload.len() >= 4 {
```
AprilNEA: `sync_clock` treats a `MSG_EXIT` frame with a payload shorter than 4 bytes as exit code 0, which can convert a malformed/partial response into a false success. Require exactly 4 bytes.

```suggestion
    if payload.len() != 4 {
        return Err(VmmError::Vsock(format!(
            "clock sync: malformed exit payload length {}, expected 4 bytes",
            payload.len()
        )));
    }
    let code = i32::from_le_bytes(payload[..4].try_into().unwrap());
```

PR #66, opened 2026-03-14
fix(net): MpmcRing soundness + reduce unwrap() in non-test code

AprilNEA:
## Summary

Based on `chore/safety-comments` (#65).

- **Fix MpmcRing soundness hole**: `MpmcRing::dequeue()` performed `assume_init_read()` before CAS succeeded — on failure the read value was discarded, which is UB for non-Copy types (double-drop). Fixed by adding `T: Copy` bound to the struct. All existing callers already use `u32`/`u64`.
- **Replace all `.lock().unwrap()` with `.lock().expect("name lock poisoned")`**: 51 occurrences across 9 files, so crash logs immediately identify which lock failed.
- **Replace `.read()/.write().unwrap()` with `.expect()`**: 29 occurrences across 6 files (RwLock, HashMap, path conversions).
- **Remove unused `try_enqueue`/`try_dequeue`** that were trivial aliases of `enqueue`/`dequeue`.

**unwrap() in non-test code: 1572 → 1201 (-371, -24%)**

Most remaining unwraps are in test-adjacent code or files where the non-test code already uses `?` properly (`arcbox-oci`, `arcbox-docker`, `arcbox-fs`).

## Test plan

- [x] `cargo check` passes
- [ ] `cargo test -p arcbox-net` passes (MpmcRing tests)
- [ ] Full `cargo test` passes

### DIFF
--- a/virt/arcbox-hypervisor/src/linux/vm.rs
+++ b/virt/arcbox-hypervisor/src/linux/vm.rs
@@ -255,7 +255,7 @@ impl KvmVm {
 
     /// Returns the current VM state.
     pub fn state(&self) -> VmState {
-        *self.state.read().unwrap()
+        *self.state.read().expect("VM state lock poisoned")
     }
 
     /// Returns whether the VM is running.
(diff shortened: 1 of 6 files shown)


AprilNEA:
Superseded by #124 — cherry-picked the MpmcRing soundness fix onto master. The unwrap→expect portion needs a fresh pass since many files have changed.

PR #67, opened 2026-03-15
fix(log): clean up guest console log formatting"

lucas77778:
- Disable ANSI color output on serial console to avoid garbled log lines
- Clean up guest console log formatting for better readability

## Test plan
- [x] Verify serial console output no longer contains ANSI escape sequences
- [x] Verify guest console logs are cleanly formatted

### DIFF
--- a/virt/arcbox-hypervisor/src/darwin/vm.rs
+++ b/virt/arcbox-hypervisor/src/darwin/vm.rs
@@ -362,7 +362,11 @@ impl DarwinVm {
                 return Ok(String::new());
             }
 
-            libc::fcntl(read_fd, libc::F_SETFL, flags | libc::O_NONBLOCK);
+            if libc::fcntl(read_fd, libc::F_SETFL, flags | libc::O_NONBLOCK) == -1 {
+                let errno = *libc::__error();
+                tracing::warn!("fcntl F_SETFL failed on fd {}: errno={}", read_fd, errno);
+                return Ok(String::new());
+            }
 
             let mut buffer = vec![0u8; 4096];
             let mut output = String::new();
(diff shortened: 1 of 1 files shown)


app/arcbox-core/src/vm_lifecycle.rs
```diff
@@ -901,20 +901,51 @@ impl VmLifecycleManager {
...
+                                                    continue;
+                                                }
+                                                line_buf.push_str(trimmed);
+                                                // Guard against runaway output without newlines.
+                                                if line_buf.len() > MAX_LINE_BUF {
+                                                    tracing::warn!(
+                                                        "Guest console line buffer overflow, flushing"
                                                     );
+                                                    line_buf.clear();
+                                                    continue;
```
AprilNEA: Bug: `continue` here skips the `tokio::time::sleep(Duration::from_millis(200))` at the bottom of the loop (line 953). If the guest streams data without newlines (e.g. binary output or a progress bar), this becomes a tight loop:

```
warn → clear → read → warn → clear → read → ...
```

Fix: add a sleep before the `continue`, or restructure so sleep is unconditional.

app/arcbox-core/src/vm_lifecycle.rs
```diff
@@ -901,20 +901,51 @@ impl VmLifecycleManager {
...
+                                                    let line =
+                                                        line_buf[..pos].trim_end().to_owned();
+                                                    line_buf.drain(..=pos);
+                                                    if line.is_empty() {
+                                                        continue;
+                                                    }
+                                                    // Agent tracing lines already have
+                                                    // timestamps — log at debug to avoid
+                                                    // double-wrapping.
+                                                    if line.contains("arcbox_agent::") {
```
AprilNEA: Two issues:

1. **Fragile matching**: `line.contains("arcbox_agent::")` depends on tracing's default format including the module path. If the crate is renamed, or tracing format changes, this breaks silently.

2. **Overly broad downgrade**: This demotes *all* agent output to `debug`, including errors and warnings. Under default `RUST_LOG=arcbox=info`, agent-side failures become invisible.

**Suggested fix**: Use a second VirtIO console port to physically separate the streams instead of content-based classification:

- `hvc0` → kernel/init output → host logs at `info`, no classification needed
- `hvc1` → agent tracing → host forwards at appropriate level

`Virtualization.framework` natively supports multiple serial ports (`setSerialPorts:` takes `NSArray`), and `add_serial_port()` already pushes to a `Vec`. The agent would open `/dev/hvc1` as its tracing writer instead of stderr. This eliminates all content-based parsing.

I'll implement this approach in follow-up commits.

PR #135, opened 2026-03-28
fix(sandbox): enable DNS resolution inside sandboxes

lucas77778:
## Summary
- Write `/etc/resolv.conf` (`nameserver 172.20.0.1`) on vm-agent startup so that DNS resolution works inside sandbox microVMs
- The sandbox kernel `ip=` boot parameter only configures IP/gateway/netmask but not DNS — without resolv.conf, glibc's resolver fails immediately without sending any query
- DNS queries now route through the guest VM's datapath `DnsForwarder` via the sandbox gateway

## Test plan
- [x] Start a sandbox and verify `ping google.com` resolves and succeeds
- [x] Verify `cat /etc/resolv.conf` inside sandbox shows `nameserver 172.20.0.1`
- [ ] Verify sandboxes with `network: none` mode still boot correctly (write failure is logged, not fatal)

### DIFF
--- a/virt/arcbox-vm/src/bin/vm-agent.rs
+++ b/virt/arcbox-vm/src/bin/vm-agent.rs
@@ -60,7 +60,7 @@ mod agent {
     const MSG_EXIT: u8 = 0x12;
 
     // File I/O channel (vsock:53) — imported from the shared proto module.
-    use arcbox_vm::KernelIpParam;
+    use arcbox_vm::boot_proto::KernelIpParam;
     use arcbox_vm::file_io::proto::{
         FILE_ACK, FILE_DATA, FILE_DONE, FILE_ERR, FILE_PORT, FILE_READ_REQ, FILE_WRITE_REQ,
         MAX_FILE_SIZE,
--- a/virt/arcbox-vm/src/lib.rs
+++ b/virt/arcbox-vm/src/lib.rs
@@ -23,7 +23,6 @@ pub mod vsock;
 pub mod instance;
 pub mod manager;
 
-pub use boot_proto::KernelIpParam;
 pub use config::{DefaultVmConfig, FirecrackerConfig, GrpcConfig, NetworkConfig, VmmConfig};
 pub use error::{Result, VmmError};
 pub use network::{NetworkAllocation, NetworkManager};


virt/arcbox-vm/src/bin/vm-agent.rs
```diff
@@ -710,8 +710,19 @@ mod agent {
         let _ = std::os::unix::fs::symlink("/dev/pts/ptmx", "/dev/ptmx");
     }
 
+    /// Write `/etc/resolv.conf` pointing at the sandbox gateway so that
+    /// userspace DNS resolution works.  The gateway runs inside the host
+    /// VM where the datapath `DnsForwarder` handles queries.
+    fn setup_dns() {
+        match std::fs::write("/etc/resolv.conf", "nameserver 172.20.0.1\n") {
```
AprilNEA: **Hardcoded gateway IP.** `172.20.0.1` is the *default* gateway, but it's configurable via `NetworkConfig::gateway` in `virt/arcbox-vm/src/config.rs`. If someone changes the gateway config, sandbox DNS silently breaks.

The gateway is already passed into the sandbox kernel via the `ip=` boot parameter (`sandbox.rs:1402`):
```
ip=<client>::<gw>:<mask>::eth0:off
```

Parse `/proc/cmdline` at startup to extract the gateway and use it as the nameserver — no magic number needed.

virt/arcbox-vm/src/bin/vm-agent.rs
```diff
@@ -710,8 +710,19 @@ mod agent {
...
+            Ok(()) => eprintln!("vmm-guest-agent: wrote /etc/resolv.conf"),
+            Err(e) => eprintln!("vmm-guest-agent: failed to write /etc/resolv.conf: {e}"),
+        }
+    }
+
     pub fn run() {
         mount_filesystems();
```
AprilNEA, line 755: **Unconditional overwrite.** `setup_dns()` runs regardless of context:
- `network: none` sandboxes — no network interface exists, but we still overwrite resolv.conf. The current error handling (log + continue) masks this, but it's conceptually wrong: the write can succeed (rootfs is mounted read-write) even when there's no network.
- Custom rootfs — if a rootfs ships its own `/etc/resolv.conf` (e.g., pointing at a local resolver), we silently clobber it.

Suggested: gate this on the `ip=` parameter being present in `/proc/cmdline` — if it's there, we have network config and can derive the nameserver; if not, leave resolv.conf alone.

virt/arcbox-vm/src/bin/vm-agent.rs
```diff
@@ -710,8 +710,19 @@ mod agent {
         let _ = std::os::unix::fs::symlink("/dev/pts/ptmx", "/dev/ptmx");
     }
 
+    /// Write `/etc/resolv.conf` pointing at the sandbox gateway so that
+    /// userspace DNS resolution works.  The gateway runs inside the host
```
AprilNEA: Nit: the doc comment says "the datapath `DnsForwarder` handles queries" — that's the Darwin outer-VMM network path (`virt/arcbox-net/src/darwin/datapath_loop.rs`), not what handles sandbox DNS. The sandbox's `172.20.0.1` hits the guest DNS server (`guest/arcbox-agent/src/dns_server.rs`) listening on `0.0.0.0:53`. These are different layers.

AprilNEA (changes requested):
The fix is on the right track — sandbox microVMs do need `/etc/resolv.conf` for glibc resolution. Two issues with the current approach:

virt/arcbox-vm/src/bin/vm-agent.rs
AprilNEA: Addressed in 9410603 and further improved in dad95bc.

Beyond parsing from `/proc/cmdline`, I also extracted the `ip=` format into a shared `KernelIpParam` type (`boot_proto.rs`) with `Display`/`FromStr`. This way `sandbox.rs` constructs via `format!("{ip_param}")` and `vm-agent` parses via `token.parse::<KernelIpParam>()` — the encoding is defined once, and any format drift becomes a compile-time or test failure instead of a silent runtime mismatch.

Placed it in `boot_proto` (not `config`) because this is a host↔guest protocol type, not host-side VMM configuration — same layer as `file_io::proto`.

virt/arcbox-vm/src/bin/vm-agent.rs
AprilNEA: Addressed. `setup_dns()` now exits early when no `ip=` token is found in `/proc/cmdline`, so `network: none` sandboxes and custom rootfs are left untouched.

virt/arcbox-vm/src/bin/vm-agent.rs
AprilNEA: Fixed. Doc comment now correctly describes the path: sandbox → `172.20.0.1` → guest DNS server (`0.0.0.0:53` in `dns_server.rs`) → upstream. No mention of `DnsForwarder`.

virt/arcbox-vm/src/lib.rs
```diff
@@ -22,6 +23,7 @@ pub mod vsock;
 pub mod instance;
 pub mod manager;
 
+pub use boot_proto::KernelIpParam;
```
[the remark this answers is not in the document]
AprilNEA: Good call. Removed the root re-export in 9bbb886 — now imported via `arcbox_vm::boot_proto::KernelIpParam` only where needed.

PR #150, opened 2026-03-31
fix(daemon): add ExitTimeOut to launchd plists to prevent SIGKILL during shutdown

AprilNEA:
## Summary
- Add `ExitTimeOut` of 45 seconds to both launchd plist files
- Prevents launchd from SIGKILL-ing the daemon during graceful shutdown (default 20s timeout is too short for 5s gRPC drain + 30s VM stop)

## Test plan
- [ ] Verify `arcbox daemon stop` completes gracefully without SIGKILL
- [ ] Verify macOS system restart doesn't leave stale sockets

Closes ABX-237


AprilNEA:
@pullfrog review

PR #151, opened 2026-03-31
fix(core): eliminate TOCTOU in MachineManager::create by holding write lock

AprilNEA:
## Summary
- Take write lock at the start of `create()` instead of read-then-write
- Existence check and insert are now atomic with respect to other writers
- Prevents concurrent creates from duplicating VMs and orphaning filesystem directories

## Test plan
- [ ] Concurrent `arcbox machine create` from CLI + desktop doesn't duplicate

Closes ABX-230

### DIFF
--- a/app/arcbox-core/src/machine.rs
+++ b/app/arcbox-core/src/machine.rs
@@ -314,13 +314,12 @@ impl MachineManager {
     ///
     /// Returns an error if the machine cannot be created.
     pub async fn create(&self, config: MachineConfig) -> Result<String> {
-        // Check if machine already exists
-        if self
-            .machines
-            .read()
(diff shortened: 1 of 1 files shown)


AprilNEA:
Closing as already-merged.

- The fix in this PR landed via 42f3e3f (identical diff, different SHA — likely a rebase artifact during a prior merge attempt).
- Copilot's first review comment (lock held across slow I/O) was addressed by 8b8ac58, which expanded the comment to record why we accept the trade-off (`create` is rare and user-driven; the `Creating`-sentinel alternative has a worse orphan-state failure mode).
- Copilot's second review comment (no concurrency regression test) is addressed in follow-up PR #289.

PR #153, opened 2026-03-31
fix(guest): bound thread spawns and reduce stack size in vm-agent

AprilNEA:
## Summary
- Add `MAX_ACTIVE_CONNECTIONS` (64) semaphore to cap concurrent connection threads
- Reduce thread stack from 8MB to 1MB via `thread::Builder::stack_size`
- Prevents guest OOM under exec bursts (e.g., docker health checks)

## Test plan
- [ ] `docker exec` burst doesn't OOM the guest

Closes ABX-238

### DIFF
--- a/virt/arcbox-vm/src/bin/vm-agent.rs
+++ b/virt/arcbox-vm/src/bin/vm-agent.rs
@@ -38,6 +38,7 @@ mod agent {
     use std::collections::HashMap;
     use std::io::{Read, Write};
     use std::os::unix::io::RawFd;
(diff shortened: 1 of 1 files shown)


AprilNEA:
Rebased onto new master (post #259/#260) in 34abc8a. Trivial rebase — touches `virt/arcbox-vm/src/bin/vm-agent.rs` which wasn't touched by the HV-parity stack. Build + clippy pass; the fix (64-conn semaphore + 1 MB stack) is unchanged and still addresses the guest OOM symptom from ABX-238.

PR #154, opened 2026-03-31
fix(net): parse and store DNS response records in cache

AprilNEA:
## Summary
- Parse answer RRs from raw DNS response bytes instead of storing empty `Vec::new()`
- Extract record type, class, TTL, and rdata (A/AAAA as IpAddr, others as raw bytes)
- Use minimum TTL from answer records for cache expiry
- Skip caching responses with no answer records (NXDOMAIN, SERVFAIL)
- DNS cache is now functional — cache hits return real data

## Test plan
- [x] `test_cache_response_stores_records` — verifies parsed records
- [x] `test_cache_hit_returns_valid_response` — full round-trip cache store → lookup
- [x] `test_cache_skips_empty_response` — NXDOMAIN not cached
- [x] All 43 DNS tests pass

Closes ABX-229

### DIFF
--- a/virt/arcbox-net/src/dns.rs
+++ b/virt/arcbox-net/src/dns.rs
@@ -617,16 +617,179 @@ impl DnsForwarder {
         Err(NetError::Dns("all upstream servers failed".to_string()))
     }
 
-    /// Caches a DNS response.
-    fn cache_response(&mut self, name: &str, qtype: DnsRecordType, _response: &[u8]) {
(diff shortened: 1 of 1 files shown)


AprilNEA:
Closing this older implementation. We'll replace it with a more structural DNS cache fix that caches validated raw upstream responses and uses a mature DNS crate for TTL extraction instead of hand-rolled packet reconstruction.

PR #155, opened 2026-03-31
fix(docker): use raw container ID as fallback for networking teardown

AprilNEA:
## Summary
- Teardown handlers (`stop_container`, `kill_container`, `remove_container`) now use `resolve_or_raw_for_teardown`, which falls back to the raw URI-extracted container ID when canonical resolution fails. This prevents port-forwarding listeners and DNS entries from leaking when the canonical-inspect call has a transient failure but the proxied lifecycle call still succeeds.
- `resolve_canonical_from_uri` keeps its strict semantics (returns `None` on failure). `rename_container` deliberately stays on the strict path: the raw URI token is often a name, and a name becomes stale after a successful rename, which would break the post-rename `inspect → register_dns` step.
- A non-matching key against the canonical-keyed network maps is a no-op — strictly better than skipping teardown entirely.

## Scope
This fixes the case where canonical resolution fails but the lifecycle proxy still succeeds (e.g. dockerd flap between the inspect and the lifecycle call). When the proxy itself fails, teardown is intentionally skipped — running it blindly could tear down an active container.

## Test plan
- [x] `cargo fmt`, `cargo clippy -p arcbox-docker -- -D warnings`, 67 unit tests + 5 integration tests pass
- [ ] Stop container while guest dockerd is restarting — verify port forwarding cleaned up
- [ ] Rename container while guest dockerd is briefly unavailable — verify DNS re-registration still happens after dockerd recovers

Closes ABX-181

### DIFF
--- a/app/arcbox-docker/src/handlers/container.rs
+++ b/app/arcbox-docker/src/handlers/container.rs
@@ -434,9 +434,9 @@ fn canonical_id_or_fallback(container_id: &str, inspect_json: &[u8]) -> String {
 }
 
 /// Extracts a container identifier from the URI and resolves it to the
-/// canonical full ID. Returns `None` (with a warning) when canonical
-/// resolution fails, so callers skip teardown rather than using a
-/// potentially wrong key (name/short-id).
+/// canonical full ID. Falls back to the raw URI-extracted ID when canonical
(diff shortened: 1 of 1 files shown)


AprilNEA:
@pullfrog review

PR #158, opened 2026-04-01
feat(sandbox): support creating sandboxes from Dockerfiles

lucas77778:
## Summary

Adds `--from-dockerfile` and `--from-image` CLI flags for creating sandboxes from Docker images, using direct overlay2 layer path resolution instead of tarball round-trips.

### Flow

```
Host CLI: --from-dockerfile → docker build → docker inspect → overlay2 layer path
Host CLI: --from-image → docker inspect → overlay2 layer path
  ↓ gRPC (rootfs = guest-internal overlay2 chain-id directory)
Guest Agent: auto-detect directory → oci2rootfs <layer-path> → ext4 → inject vm-agent → boot
```

- `--from-dockerfile` runs `docker build` (proxied to guest dockerd via ArcBox Docker context), then `docker inspect` to get the overlay2 `UpperDir` → chain-id directory
- `--from-image` directly inspects an existing image for its layer path
- `--rootfs` continues to work for direct ext4 image paths (no change)
- Guest agent auto-detects: if rootfs is a directory → convert via `oci2rootfs`; if file → use as ext4 directly
- Cached ext4 images (keyed by layer path hash) skip conversion on subsequent creates

### What changed vs previous iteration

Addressed [PeronGH's design review](https://github.com/arcboxlabs/arcbox/pull/158#pullrequestreview-2803458893):

- **No `docker save` / tarball round-trip** — `oci2rootfs` now reads overlay2 directories directly ([arcboxlabs/oci2rootfs#1](https://github.com/arcboxlabs/oci2rootfs/pull/1))
- **Removed `rootfs_type` proto field** — guest agent auto-detects via `Path::is_dir()`
- **Three explicit CLI flags** instead of auto-detecting file format on `--rootfs`
- **Removed `/images/{id}/get` and `/images/get`** Docker compat proxy routes (no longer needed)
- **No `~/.arcbox/bin/` staging** — layer path is passed directly from `docker inspect`
- **Net -127 lines** — simpler architecture, less code

### Usage

```bash
# From Dockerfile
arcbox sandbox create --from-dockerfile /path/to/Dockerfile --memory 1024

# From existing Docker image
arcbox sandbox create --from-image ubuntu:22.04 --memory 1024

# Direct ext4 (unchanged)
arcbox sandbox create --rootfs /path/to/rootfs.ext4 --memory 1024
```

## Test plan

- [x] `cargo clippy` + `cargo fmt` — zero new warnings
- [x] `cargo test` — all tests pass
- [x] `arcbox sandbox create --rootfs <ext4>` still works (no regression)
- [x] `arcbox sandbox create --from-dockerfile <Dockerfile>` builds, converts, and boots
- [x] `arcbox sandbox create --from-image <ref>` resolves, converts, and boots
- [x] Second create with same Dockerfile/image hits cache and skips conversion

### DIFF
--- a/app/arcbox-cli/src/commands/sandbox.rs
+++ b/app/arcbox-cli/src/commands/sandbox.rs
@@ -86,9 +86,15 @@ pub struct CreateArgs {
     /// Kernel image path (empty = daemon default)
     #[arg(long)]
     pub kernel: Option<String>,
-    /// Root filesystem image path (empty = daemon default)
-    #[arg(long)]
+    /// Root filesystem ext4 image path (empty = daemon default)
+    #[arg(long, conflicts_with_all = ["from_dockerfile", "from_image"])]
     pub rootfs: Option<String>,
+    /// Build sandbox rootfs from a Dockerfile
+    #[arg(long, conflicts_with_all = ["rootfs", "from_image"])]
+    pub from_dockerfile: Option<String>,
+    /// Build sandbox rootfs from an existing Docker image
+    #[arg(long, conflicts_with_all = ["rootfs", "from_dockerfile"])]
+    pub from_image: Option<String>,
     /// Number of vCPUs (0 = daemon default)
     #[arg(long, default_value = "0")]
     pub cpus: u32,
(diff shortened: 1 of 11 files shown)


PeronGH (commented):
## Design feedback

The guest agent should never interact with Docker — it only boots Firecracker microVMs from ext4 images. Image conversion is handled by `oci2rootfs`, a dedicated binary bundled in the guest VM's boot assets that takes an explicit OCI layer path as input.

### Proposed design

Three CLI flags converging through shared code paths:

```mermaid
flowchart LR
    subgraph Host CLI
        A["--from-dockerfile"] --> B["docker build\n(→ guest dockerd)"]
        B --> C["image ref"]
        D["--from-image"] --> C
        C --> E["docker inspect\n→ layer path"]
    end
    subgraph Guest Agent
        E --> F["exec oci2rootfs\n‹layer-path›"]
        F --> G["ext4"]
    end
    H["--rootfs"] --> G
    G --> I["boot Firecracker\nmicroVM"]
```

Since macOS `/Users` is VirtioFS-mounted at the same path in the guest, and `docker inspect` returns guest-internal paths (dockerd runs in the guest), all paths work as-is — no translation needed anywhere.

- `--from-dockerfile` runs `docker build` on the host (proxied to guest dockerd via existing Docker compat layer), then passes the resulting image ref internally to `--from-image`
- `--from-image` gets the layer path via `docker inspect`, passes it to the agent which execs `oci2rootfs <layer-path>` to produce ext4, then enters the `--rootfs` path
- `--rootfs` path is valid in both host and guest thanks to `/Users` VirtioFS passthrough
- `oci2rootfs` is a standalone binary bundled in guest boot assets — takes an explicit path, knows nothing about Docker

### What to change

- `rootfs_type` proto field is unnecessary
- No `docker save` / tarball round-trip — `oci2rootfs` reads layers directly from the path returned by `docker inspect`
- Guest-side `rootfs_builder.rs` replaces with a thin exec of `oci2rootfs`
- Host-side `rootfs_builder.rs` reduces to `docker build` (for `--from-dockerfile`) and `docker inspect` (for `--from-image`)
- No `~/.arcbox/bin/` staging
- `/images/{id}/get` and `/images/get` proxy routes are not needed

PR #186, opened 2026-04-06
test(net): add network datapath mock tests via socketpair

AprilNEA:
## Summary

- Add mock guest NIC infrastructure using `socketpair(AF_UNIX, SOCK_DGRAM)` to test the full network datapath without requiring a VM, root privileges, or code signing
- Implement L2 Ethernet frame builders (ARP, DHCP, TCP SYN, DNS, UDP, ICMP) and parsers for test assertions
- Add 5 integration tests covering DHCP full cycle and frame classification (ARP, TCP SYN gating, ICMP, DHCP interception)

Key finding: `NetworkDatapath` already accepts a plain `OwnedFd` — no refactoring was needed. The architecture is inherently test-friendly.

**Linear**: ABX-298, ABX-300

## New files

| File | Purpose |
|------|---------|
| `virt/arcbox-net/tests/helpers/mock_nic.rs` | socketpair mock NIC + FD I/O utilities |
| `virt/arcbox-net/tests/helpers/frames.rs` | L2 frame builders and parsers |
| `virt/arcbox-net/tests/datapath_test.rs` | 5 integration tests |

## Test plan

- [x] `cargo test -p arcbox-net --test datapath_test` — all 5 tests pass
- [ ] `cargo clippy -p arcbox-net` — zero warnings
- [ ] CI passes on macOS runner

### DIFF
--- a/virt/arcbox-net/tests/datapath_test.rs
+++ b/virt/arcbox-net/tests/datapath_test.rs
@@ -67,6 +67,7 @@ fn create_test_datapath() -> (NetworkDatapath, std::os::fd::OwnedFd, Cancellatio
         GUEST_IP,
         GATEWAY_MAC,
         cancel.clone(),
+        1500, // mtu
     );
 
     (datapath, guest_fd, cancel)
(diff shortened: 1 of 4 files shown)


AprilNEA:
Rebased onto new master (post #259/#260) in 73caa4c. Required one adjustment: `NetworkDatapath::new` and `SmoltcpDevice::new` now take an explicit `mtu: usize` parameter, so the mock datapath harness passes `1500` (the default MTU used in production paths). All 5 tests (DHCP full cycle + ARP / TCP SYN / ICMP / DHCP classification) pass locally.

Worth noting: the smoltcp removal did not invalidate these tests. The frame classifier still recognizes ARP/DHCP/ICMP the same way, and the TCP SYN test was already asserting "not intercepted" which is still true — TCP SYN now flows through the hand-rolled handshake synthesizer instead of smoltcp's socket, but from the classifier's perspective the non-interception assertion holds.

PR #203, opened 2026-04-06
perf(net): TCP fast path — bypass smoltcp for established connections

AprilNEA:
## Summary

Add a TCP fast path that bypasses smoltcp's per-segment TCP state machine for established connections, directly relaying TCP payload between guest frames and host TcpStreams.

### Architecture

```
Guest TCP segment → classify_ipv4() → 5-tuple lookup
  → HIT: extract payload → write to host TcpStream → generate 54-byte ACK → guest FD
  → MISS: normal smoltcp path (unchanged)

Host TcpStream readable → poll_fast_path()
  → read data → build ETH+IP+TCP frame (checksums) → guest FD
```

smoltcp still handles SYN gating, 3-way handshake, and FIN/RST teardown. Only data segments on established connections use the fast path.

### Why this works without full TCP

The transport between the fast path and the guest is VirtIO-net over AF_UNIX socketpair — reliable, in-order, no packet loss. We only need:
- Correct SEQ/ACK numbers (simple counters)
- Valid IP + TCP checksums (reuse existing utilities)
- FIN/RST detection for teardown (fall through to smoltcp)

### Changes

| File | Change |
|------|--------|
| `ethernet.rs` | `TcpFrameParams`, `tcp_checksum()`, `build_tcp_ack/data/fin/rst_frame()` |
| `tcp_bridge.rs` | `FastPathConn`, `try_fast_path_intercept()`, `poll_fast_path()`, `promote_to_fast_path()` |
| `smoltcp_device.rs` | `drain_fast_path()` — filters rx_queue before smoltcp sees frames |
| `datapath_loop.rs` | Wire drain_fast_path + poll_fast_path into event loop |

### Status

The fast-path infrastructure is complete and wired in. Connections must be explicitly promoted via `promote_to_fast_path()`. A follow-up will add automatic promotion in `detect_new_connections()`.

## Test plan

- [x] `cargo test -p arcbox-net` — 228 passed, 0 failed
- [x] `cargo clippy -p arcbox-net --lib` — zero errors/warnings
- [ ] E2E: `docker run -p 8080:80 nginx` + `curl localhost:8080`
- [ ] Benchmark: iperf3 with promoted fast-path connections

### DIFF
--- a/virt/arcbox-net/src/darwin/datapath_loop.rs
+++ b/virt/arcbox-net/src/darwin/datapath_loop.rs
@@ -275,7 +275,18 @@ impl NetworkDatapath {
                     if prev_mac.is_none() {
                         if let Some(gmac) = guest_mac {
                             device.seed_gateway_neighbor(gateway_ip, gateway_mac, gmac);
+                            tcp_bridge.set_fast_path_macs(gateway_mac, gmac);
                         }
+                    }
+
+                    // Fast-path intercept: extract TCP data frames for established
+                    // fast-path connections BEFORE smoltcp sees them. This avoids
+                    // smoltcp's per-segment TCP state machine overhead.
+                    let fast_acks = device.drain_fast_path(|frame_data| {
+                        tcp_bridge.try_fast_path_intercept(frame_data)
+                    });
+                    for ack in fast_acks {
+                        enqueue_or_write(&guest_async, FrameBuf::from(ack), &mut write_queue);
                     }
 
                     // Process intercepted frames (DHCP, DNS, UDP, ICMP).
@@ -409,7 +420,13 @@ impl NetworkDatapath {
             let ts = smoltcp::time::Instant::now();
             iface.poll(ts, &mut device, &mut sockets);
 
-            // 5. Flush TX frames to guest.
+            // 5. Poll fast-path host streams for inbound data and inject
+            //    constructed frames directly to guest (bypasses smoltcp).
+            for frame in tcp_bridge.poll_fast_path() {
+                enqueue_or_write(&guest_async, FrameBuf::from(frame), &mut write_queue);
+            }
+
+            // 6. Flush TX frames to guest (from smoltcp).
             for frame in device.take_tx_pending() {
                 enqueue_or_write(&guest_async, frame, &mut write_queue);
             }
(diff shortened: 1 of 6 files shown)


AprilNEA:
Addressed the three real bugs from review in 6520ef7:

1. **WouldBlock no longer kills connection** — non-blocking write returning WouldBlock now skips ACK (guest retransmits). Connection preserved.

2. **Host→guest segmented to MSS** — reads are chunked into 3960-byte segments (MTU 4000 - IP(20) - TCP(20)) instead of a single 32 KiB frame that would exceed MTU.

3. **Payload bounded by IPv4 total_length** — excludes Ethernet padding bytes that could corrupt the relay.

Also optimized `drain_fast_path()` to filter in-place instead of rebuilding VecDeque.

Regarding the repeated comments on net_mtu hardcoding and PacketPool size — these were already discussed and resolved in #198 review. Our target is macOS 14+ and the 256 MiB pool is intentional for burst absorption.

AprilNEA:
Second-round review comments addressed in 463ee01 + 30aba27:

**Fixed:**
- `into_std()` failure no longer panics — graceful error logging instead of `unwrap()` fallback
- Removed unnecessary `unsafe impl Send` for `FrameBuf` — replaced with compile-time const assertion

**Clarified with comments:**
- MTU hardcoding: added `NOTE(MTU)` in darwin.rs explaining why 4000 is intentional for macOS 14+ target and under what conditions mismatch matters
- PacketPool 256 MiB: existing comment already explains the rationale (one per VM, idle pages compressed, smaller pool defeats purpose)

**Already correct (no change needed):**
- Partial write: ACKs only `n` bytes written, guest retransmits remainder — correct TCP behavior
- `payload_start` bounds: already protected by `saturating_sub` from previous fix
- MSS constant 3960: matches our MTU 4000 target; non-target platforms are documented as unsupported

AprilNEA:
Third-round review addressed in 1b6a50c:

**Fixed (real bugs):**
- **FIN/RST teardown**: promoted connections now handle teardown themselves — guest FIN gets FIN+ACK reply, guest RST closes host stream. Previously these fell through to smoltcp which had no matching socket.
- **Host EOF**: fast-path entry stays alive after sending FIN so guest's FIN-ACK is properly handled.

**Fixed (comments/typos):**
- darwin.rs NOTE(MTU): corrected `<13` → `<14`
- into_std failure: log no longer claims smoltcp fallback (connection is lost)
- drain_fast_path: comment matches implementation

**Noise (no change):**
- MSS 3960 hardcode: matches MTU 4000 target, documented
- Guest window tracking: VirtIO backpressure + write_queue cap is sufficient
- Test coverage: valid suggestion, deferred to follow-up
- PacketPool 256 MiB: discussed 4 times now, intentional

PR #204, opened 2026-04-06
fix(net): defer fast-path promotion to first data frame

AprilNEA:
## Problem

PR #203 introduced fast-path promotion in `relay_all()` that caused connection resets. When smoltcp received the final ACK and first data frame (e.g. TLS ClientHello) in the same batch, the data was buffered in smoltcp's socket. Promotion then aborted the socket, destroying the buffered data. Guest never got a response → RST.

This broke Docker image pulls through proxy tunnels (Surge/Clash fake-IP).

## Fix

Defer promotion from `relay_all()` to `try_fast_path_intercept()`:

```
1. detect_new_connections(): store pre-connected stream in held_streams
   + spawn host_conn_task for relay during handshake
2. relay_all(): ESTABLISHED → mark in fast_path_ready (don't promote)
3. try_fast_path_intercept(): first DATA frame → inline promote
   (SEQ/ACK from frame headers, stream from held_streams)
```

The first data frame triggers promotion in `drain_fast_path()` BEFORE `iface.poll()`, so smoltcp never sees it → no data loss.

## Scope

Fast path currently activates for **outbound** connections with pre-connected streams (SYN gate). Inbound port forwarding (`initiate_inbound`) is not yet covered — follow-up needed.

## Test plan

- [x] `cargo test -p arcbox-net` — 228 passed
- [x] `docker run --rm alpine echo hello` → hello
- [x] Docker image pull through Surge proxy → works
- [x] iperf3 benchmark: ~2.3 Gbps (baseline, fast path not triggered for inbound)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ tests/resources/boot-assets-arm64-*
 # Scratch notes
 perf-issue.md
 issues.net.md
+docs/audit.md
--- a/docs/audit.md
+++ b/docs/audit.md
@@ -1,0 +1,929 @@
+# Codebase Audit Report
+
+> Date: 2026-03-29 (updated)  
+> Scope: full workspace (327 Rust files, ~110K LOC)  
+> Branch: `fix/daemon-shutdown-feedback`  
+> Version: `v0.3.13`  
+> Method: merged findings from three independent full-repo audits plus direct code verification
+
+## Executive Summary
+
+The repo still has strong macro-level layering: no obvious circular dependencies,
+clear crate families, good use of `CancellationToken`, and a mostly coherent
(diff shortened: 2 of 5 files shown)


AprilNEA:
Superseded by #259 — the entire subsystem this PR fixed (smoltcp `relay_all()` promoting sockets and dropping buffered data) no longer exists on master. The new architecture lands smoltcp's removal and a hand-rolled TCP handshake synthesizer that holds the host stream in `HandshakeConn` until the guest ACK arrives, then promotes via `promote_to_fast_path` with SEQ/ACK captured directly from the frame headers — exactly the defer-to-first-data semantics this PR was aiming for. Closing as obsolete.

PR #208, opened 2026-04-07
feat(vm): dm-snapshot CoW with jailer mode support

lucas77778:
## Summary

- Add dm-snapshot based block-level CoW for sandbox rootfs, eliminating full rootfs copies on sandbox creation
- Wire dm-snapshot into the sandbox lifecycle with automatic teardown on stop/remove
- Support jailer mode by creating block device nodes (mknod) in the jailer chroot pointing to dm-snapshot devices
- Fix Firecracker jailer panic caused by missing CPU cache topology in Virtualization.framework guest sysfs (synthesise `size`, `coherency_line_size`, `number_of_sets` via bind mounts)
- Mount `/var` without `nodev` so jailer chroot block device nodes are accessible
- Enable jailer as the default sandbox execution mode

## Performance

Single sandbox (1GB rootfs, 1536MB memory): **0.6s** to ready  
3 concurrent sandboxes (same config): **1.36s** all ready  
Guest-side boot time (jailer spawn → FC config → microVM ready): **~13ms**

## Test plan

- [x] Verified missing sysfs files in guest VM via `docker run --privileged --pid=host`
- [x] Confirmed jailer panic on missing `/sys/devices/system/cpu/cpu0/cache/index0/size`
- [x] Confirmed cache fixup bind mounts applied correctly after agent restart
- [x] Confirmed `/var` tmpfs `nodev` was root cause of block device Permission denied
- [x] Created and ran sandbox in jailer mode successfully (`sandbox run -- uname -a`)
- [x] Benchmarked 1 and 3 concurrent sandbox creation with 1GB rootfs


AprilNEA:
@pullfrog review

PR #209, opened 2026-04-08
feat(vm): symlink indirection for dm-snapshot checkpoint/restore

lucas77778:
## Summary

- Add symlink indirection layer between Firecracker and dm-snapshot devices, inspired by E2B's architecture. FC's vmstate records a stable symlink path (`{vm_dir}/rootfs.link`) instead of the ephemeral `/dev/mapper/arcbox-snap-{id}` device path. On restore, a new dm-snapshot is created and the symlink is retargeted — FC reopens the same path transparently.
- Wire dm-snapshot CoW into `restore_sandbox()` for both jailer and direct modes, eliminating full rootfs copies on restore.
- Store `cow_handle` on restored `SandboxInstance` so `remove_sandbox_impl()` properly tears down the dm-snapshot.
- Remove unused `stage_files_for_jailer()` — all call sites now use the split `stage_kernel_for_jailer()` + `stage_rootfs_*_for_jailer()`.

### How it works

**Boot (direct mode):**
```
cow_manager.setup() → /dev/mapper/arcbox-snap-{id}
symlink: {vm_dir}/rootfs.link → /dev/mapper/arcbox-snap-{id}
FC receives: {vm_dir}/rootfs.link (stable path, saved in vmstate)
```

**Restore (direct mode):**
```
cow_manager.setup() → /dev/mapper/arcbox-snap-{new_id}
symlink: {original_vm_dir}/rootfs.link → /dev/mapper/arcbox-snap-{new_id}
FC loads vmstate → opens {original_vm_dir}/rootfs.link → new device
```

**Restore (jailer mode):**
```
cow_manager.setup() → /dev/mapper/arcbox-snap-{new_id}
mknod: {chroot}/rootfs.ext4 → (major, minor) of new dm device
FC loads vmstate → opens /rootfs.ext4 (chroot-relative) → new device
```

## Test plan

- [x] Create sandbox → checkpoint → remove → restore → verify `ready` state
- [x] Verified `dm-snapshot created` log for both original and restored sandbox
- [x] Verified `dm-snapshot teardown complete` on remove of both
- [x] `cargo clippy` and `cargo fmt` pass with zero warnings
- [ ] CI

### DIFF
--- a/virt/arcbox-vm/src/sandbox.rs
+++ b/virt/arcbox-vm/src/sandbox.rs
@@ -881,13 +881,10 @@ impl SandboxManager {
             (snap_dir.join("vmstate"), snap_dir.join("mem"))
         };
 
-        // Store kernel/rootfs only in jailer mode — needed when restoring.
-        let (snap_kernel, snap_rootfs) = if self.config.firecracker.jailer.is_some() {
-            (Some(kernel_path), Some(rootfs_path))
-        } else {
-            (None, None)
-        };
-
+        // Store kernel/rootfs template paths so restore can re-derive them.
+        // Jailer mode needs them for chroot staging; direct mode needs the
+        // rootfs path to set up a fresh dm-snapshot and retarget the
+        // vmstate-recorded symlink.
         let meta = self.snapshots.register(
             sandbox_id,
             Some(name),
@@ -895,8 +892,8 @@ impl SandboxManager {
             vmstate_path,
             Some(mem_path),
             None,
-            snap_kernel,
-            snap_rootfs,
+            Some(kernel_path),
+            Some(rootfs_path),
         )?;
 
         let snap_dir_path = meta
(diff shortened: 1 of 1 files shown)


AprilNEA:
@greptile review

AprilNEA:
Rebased onto master (PR 208's later evolution made the branch's 7 stale PR-208 commits unrebaseable). Force-pushed as a single coherent commit. Old commits archived at tag `backup/pr-209-pre-reset`.

Addressed all three greptile findings:

**P1 — CowHandle leak on symlink failure (restore):** the symlink call in direct-mode restore is now wrapped in a `match`. On `Err`, `cow_manager.teardown(&handle).await` runs before the error propagates, releasing the dm device, loop device, and COW file.

**P1 — `original_vm_dir` leaked after restored sandbox removal:** added `restore_origin_dir: Option<PathBuf>` to `SandboxInstance`, populated during direct-mode restore. `remove_sandbox_impl` removes it (alongside the new sandbox's `vm_dir`) so each restore-and-remove cycle cleans up after itself.

**P2 — Dangling `rootfs.link` on boot failure:** `do_boot`'s post-`builder.start()` failure path now removes `vm_dir/rootfs.link` after `cow_manager.teardown`, gated to direct mode (jailer uses a chroot-internal device node that disappears with the chroot).

Also applied the same teardown guard to the boot-time symlink path: if `create_rootfs_symlink` fails after `cow_manager.setup` succeeds, the handle is torn down before propagation.

AprilNEA:
@pullfrog review

AprilNEA:
Addressed pullfrog review on `bab5290`.

**Critical: direct-mode restore was non-functional.** `checkpoint_sandbox` gated `rootfs_path`/`kernel_path` storage on `jailer.is_some()`, so direct-mode snapshots had `snap_meta.rootfs_path = None`. The new restore CoW branch was dead code in direct mode, and FC would fault opening the vmstate-recorded `rootfs.link` against the empty dir we just recreated. Dropped the gate — both paths are now always stored.

**CowHandle leak across post-setup failure paths.** The previous P1 patch only covered symlink-creation failure. Refactored so `pending_cow` and `pending_origin_dir` live across both setup branches; every failure path (jailer vmstate/mem copy + chown, `fc_sdk::restore` in either mode) routes through a new `cleanup_pending_restore` helper that tears the handle down (and removes the recreated origin dir for direct mode) before propagating.

**Cosmetic.** Replaced `link_path.to_str().unwrap()` with a `VmmError::Config` on non-UTF-8 paths, and reworded the symlink-failure comment to drop the "P1 fix:" review annotation.

AprilNEA:
Addressed the latest pullfrog + greptile pass on `c265db7`.

**greptile P1: `dmsetup remove` race with FC.** `FirecrackerProcess::drop` SIGKILLs but never reaps, so teardown right after dropping FC could see EBUSY (FC still has the dm device open) and silently leak the dm + loop + COW resources. New `kill_and_reap_fc` helper sends SIGKILL and `wait()`s with a 5s timeout before `cow_manager.teardown`. Wired into both `restore_sandbox` failure paths and `do_boot`'s `builder.start()` failure path.

**pullfrog: `pending_origin_dir` only tracked on CoW Ok branch.** The directory is created right after the vsock-path setup, but `pending_origin_dir` was only set inside the CoW success branch — any failure between `create_dir_all` and CoW setup leaked the orphan directory. Set as soon as the directory exists.

**pullfrog: silent direct-mode CoW fallthrough.** The `cow_manager.setup` Err arm only emitted a debug log, leaving FC to fault on the missing symlink with a generic error. Unlike boot, direct-mode restore has no safe fallback (the vmstate-recorded path is the symlink itself). Now returns an explicit `VmmError::DeviceMapper` so the failure mode is clear.

PR #267, opened 2026-04-24
fix(agent): make ensure_runtime non-blocking to avoid daemon RPC timeout

AprilNEA:
## Summary

The agent's `handle_ensure_runtime` previously awaited the full runtime-start sequence inline (up to ~90s on cold boot, polling for dockerd readiness). The host-side vsock `BlockingVsockTransport` uses a 5s deadline per RPC, so the daemon would close the socketpair mid-call and the agent's eventual response write would fail with EPIPE.

Symptoms users saw:
- Daemon spam: `guest docker endpoint on vsock port 2375 not ready within 60000ms`
- Agent log spam: `ERROR Connection error: failed to write message`
- First `docker run` after fresh daemon: red-error experience, sometimes required retry

## What changed

- **`ensure_runtime` is now spawn-and-return**: first caller transitions `NotStarted`/`Failed` -> `Starting`, spawns the start sequence as a background task, and returns `STATUS_STARTING` immediately. Concurrent callers see `Starting` and also return `STATUS_STARTING`. Once the background task publishes state -> `Ready`/`Failed`, subsequent callers get cached `STATUS_REUSED` / `STATUS_FAILED`.
- **New status constant**: `RUNTIME_STARTING = "starting"`.
- **Log downgrade**: "Connection error" -> "Connection closed by peer" at `warn!` level when the `io::Error` chain indicates peer disconnect (`BrokenPipe`/`ConnectionReset`/`ConnectionAborted`/`UnexpectedEof`) — these are routine daemon-side retry teardowns, not agent faults.
- **7 new unit tests** cover fast-path Ready, Starting-returns-quickly, Failed -> Ready retry, probe-only, probe-after-Ready, and concurrent-callers-spawn-exactly-once. All green on host target.
- Daemon-side `wait_guest_endpoint_ready` already retries on `ready=false` with backoff, so no daemon change is needed.

## Test plan

- [x] `cargo test -p arcbox-agent --lib --bins` — 33 tests pass (7 new)
- [x] `cargo build -p arcbox-agent --target aarch64-unknown-linux-musl --release` — clean
- [x] `cargo build -p arcbox-daemon -p arcbox-cli -p arcbox-helper` — clean
- [x] e2e on macOS:
  - Fresh daemon + first `docker ps` / `curl /_ping`: ~3s (dominated by vsock cold-boot, no longer 60s timeout)
  - Subsequent calls: sub-100ms
  - `docker run -d alpine sleep 30` + `docker logs` + `docker rm -f`: works
  - `docker run -d -p 8082:80 nginx:alpine` + `curl localhost:8082`: HTTP 200 OK
  - daemon.log: zero `"guest docker endpoint ... not ready"` errors
  - agent.log: zero `ERROR`-level connection errors
  - Vsock agent-port (1024) RPCs complete in <1ms each (was ~5s mid-boot)

## Known unrelated issue

`docker run` in foreground/attach mode (without `-d`) prints container stdout correctly but CLI hangs after container exit — separate docker-attach HTTP-upgrade proxy bug, to be investigated in a follow-up PR.

### DIFF
--- a/app/arcbox-cli/src/commands/cli_plugins.rs
+++ b/app/arcbox-cli/src/commands/cli_plugins.rs
@@ -1,0 +1,517 @@
+//! Docker CLI plugin registration.
+//!
+//! Makes `docker compose` and `docker buildx` (space-separated subcommands)
+//! discoverable by the upstream `docker` CLI, which looks plugins up via:
+//!
+//!   1. `cliPluginsExtraDirs` in `~/.docker/config.json`
+//!   2. `~/.docker/cli-plugins/<name>`
+//!   3. `/usr/local/lib/docker/cli-plugins/<name>`
+//!   4. `/usr/lib/docker/cli-plugins/<name>`
+//!
+//! We wire up both (1) and (2) for belt-and-suspenders coverage:
+//!
+//! - **(2) symlinks** are the idiomatic registration mechanism and are what
+//!   most users expect to see when auditing their Docker setup.
+//! - **(1) extraDirs** serves as a fallback if a user wipes
+//!   `~/.docker/cli-plugins/` or hasn't got it at all (it doesn't exist on
+//!   a fresh machine without Docker Desktop installed).
+//!
+//! Both point at `~/.arcbox/bin/<name>`, which `setup::install()` has
+//! already populated with symlinks into the app bundle or runtime bin.
+//!
+//! Plugin registration is per-user and decoupled from Docker context
+//! `enable`/`disable`. Compose/buildx themselves are context-aware (they
+//! honour `DOCKER_HOST` / the current context), so the binaries work
+//! correctly even when the user has switched to another Docker backend.
+//! This means `docker compose` keeps working after `abctl docker disable`
+//! — it just talks to whichever backend the user switched to.
+
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result};
+use serde::Serialize;
+
+use arcbox_constants::paths::DOCKER_CLI_PLUGINS;
+
+/// Summary of what a register/unregister call did.
+#[derive(Debug, Default, Serialize)]
+pub struct Outcome {
+    /// Plugin symlinks created or removed under `~/.docker/cli-plugins/`.
+    pub symlinks: Vec<PathBuf>,
+    /// Whether `cliPluginsExtraDirs` in `~/.docker/config.json` was modified.
+    pub config_updated: bool,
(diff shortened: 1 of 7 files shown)


AprilNEA:
Addressed review comments in 509ef239. Summary per comment:

**ensure_runtime.rs**

- **P1 — Background task panic leaves state stuck at Starting** (greptile L155, Copilot L157): Wrapped the spawned future in `AssertUnwindSafe(fut).catch_unwind()`. A panic converts to a `Failed` state with the panic message, logged at `error!`. Added `panic_in_start_task_transitions_to_failed` test verifying the path and the subsequent retry transition.
- **probe_fn is eager while make_start is lazy** (greptile L101): Changed `probe_fn` to `FnOnce() -> Fut`, symmetric with `make_start`. Updated the call site in `agent/mod.rs` and all tests.
- **Docs vs logic: Failed → retry behavior** (Copilot L8): Retry-on-Failed is the intended behavior — the daemon's poll loop asks the agent to try again. Updated the module doc to describe this accurately: `Failed` callers transition back to `Starting` on `start_if_needed=true`; they don't observe a cached `STATUS_FAILED`. Probe-only callers re-run `probe_fn`.
- **runtime_guard doc about 'static** (Copilot L71): Adopted the suggested wording. The reason for `Arc` is "cheap clone + owned move into the spawn," not avoidance of `'static` bounds (which `tokio::spawn` still requires).

**cli_plugins.rs**

- **outcome.symlinks.push runs on non-Unix** (greptile L124): Moved both the `symlink()` call and the `symlinks.push()` into the same `#[cfg(unix)]` block.
- **register() always mutates config.json** (Copilot L125): Track `any_plugin_present` during the symlink loop; only call `update_extra_dirs(..., insert=true)` when at least one plugin binary exists. Added `register_without_plugin_binaries_leaves_config_untouched` test.
- **Precedence docs vs behavior** (Copilot L71): Docker searches `cliPluginsExtraDirs` before `~/.docker/cli-plugins/`, so registering `user_bin` there makes ArcBox win over a foreign symlink. Rewrote the doc to reflect this and explain why it's the right default: users installing ArcBox almost certainly want ArcBox's compose/buildx (pointing at ArcBox's socket). `unregister()` removes the entry so Desktop becomes selectable again on uninstall.

All tests pass: `cargo test -p arcbox-agent --lib --bins` → 34 passed, `cargo test -p arcbox-cli cli_plugins::` → 11 passed. Linux target build clean.

PR #273, opened 2026-04-24
fix(storage): increase data disk to 8 TiB and add btrfs auto-resize

AprilNEA:
## Summary

The default `docker.img` sparse file is 64 GiB, which users hit too easily once they keep a working set of images and volumes. OrbStack ships 8 TiB. Since the host file is sparse, the virtual size costs nothing until blocks are actually written.

Changes:

- `app/arcbox-core/src/vm_lifecycle/mod.rs` — bump `DOCKER_DATA_IMAGE_SIZE_BYTES` from 64 GiB to 8 TiB, with a docstring explaining sparse semantics.
- `guest/arcbox-agent/src/agent/mod.rs`:
  - Wait up to 5 s (50 × 100 ms) for `/dev/vdb` to appear. The VirtIO block device registration races the agent on cold boot.
  - After the first mount, call `BTRFS_IOC_RESIZE` with `"max"` so an existing 64 GiB filesystem transparently grows to fill the (possibly enlarged) block device on the next boot. No-op when the FS already fills the device.

Rebased from an old `fix/data-disk-size` branch (single commit from 2026-03-05). Paths updated for the `vm_lifecycle.rs → vm_lifecycle/mod.rs` and `agent.rs → agent/mod.rs` splits; the `// =====` section divider the original introduced is intentionally dropped to match the current convention.

## Test plan

- [x] `cargo check -p arcbox-core` clean
- [ ] Cross-compile `arcbox-agent` for `aarch64-unknown-linux-musl` (rely on CI)
- [ ] E2E: upgrade a machine that had the 64 GiB image and verify the mount auto-grows on next boot (`df -h /mnt/data`)
- [ ] Fresh install: verify `docker.img` virtual size reports 8 TiB and actual disk use stays low until workload writes

## Notes

The 5 s timeout for `/dev/vdb` is a heuristic — if anyone sees it expire in practice, the underlying kernel probe is stuck and increasing the timeout would only mask the symptom.

### DIFF
--- a/app/arcbox-core/src/vm_lifecycle/mod.rs
+++ b/app/arcbox-core/src/vm_lifecycle/mod.rs
@@ -69,8 +69,13 @@ const BALLOON_SHRINK_DELAY_SECS: u64 = 10;
 
 /// Persistent guest dockerd data image name.
 const DOCKER_DATA_IMAGE_NAME: &str = "docker.img";
-/// Persistent guest dockerd data image size (64 GiB sparse file).
-const DOCKER_DATA_IMAGE_SIZE_BYTES: u64 = 64 * 1024 * 1024 * 1024;
+/// Persistent guest dockerd data image size (8 TiB sparse file).
+///
+/// This is the virtual size of the block device. The host file is sparse and
+/// only consumes actual disk space for written blocks. 8 TiB matches OrbStack
+/// and prevents users from hitting artificial limits.
+const DOCKER_DATA_IMAGE_SIZE_BYTES: u64 = 8 * 1024 * 1024 * 1024 * 1024;
+
 /// Extended VM lifecycle state.
 ///
 /// This extends the basic `MachineState` with additional states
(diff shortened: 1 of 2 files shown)


AprilNEA:
Addressed the review threads and resolved the rebase conflict against current master (the agent was split into `agent/linux/*.rs`).

**Greptile P2 — resize failures swallowed**: `btrfs_resize_max` now returns `Result<String, String>`. The success/failure note is pushed to the mount notes returned from `ensure_data_mount`, so a silent failure on the 64 GiB → 8 TiB upgrade path is visible to the host. Genuine errors also log at `error!` instead of `warn!`. Mount is not failed on resize error — running on old capacity is preferable to refusing to start.

**Copilot — APFS preallocation interacting badly with 8 TiB**: agreed. Capped `F_PREALLOCATE` at 64 GiB (`APFS_PREALLOC_CAP_BYTES`). The full virtual size still goes through `set_len` (sparse), so user-visible disk capacity is 8 TiB. The bounded preallocation preserves the original double-CoW protection for the initial working set without asking APFS to physically reserve multi-TiB.

**Copilot — `args[0] = 1` little-endian assumption**: factored out to `build_resize_args(devid: i64, size: &str)` that uses `devid.to_le_bytes()` explicitly.

**Copilot — no automated coverage for ioctl args**: added `#[cfg(test)]` unit tests for `build_resize_args` covering layout and the oversized-token guard.

AprilNEA:
@pullfrog review

PR #293, opened 2026-05-31
ABX-375: All-in-HV + FEX runtime

AprilNEA:
## ABX-375: All-in-HV + FEX64 for `linux/amd64` — status & handoff

**Goal:** run `linux/amd64` OCI containers through FEX64 (binfmt) inside the single
HV utility VM, so no VZ/Rosetta runtime VM is needed. VZ/Rosetta kept only as an
optional build backend / fallback (ABX-374, PR #291, branch
`feat/dual-utility-vm-routing` — preserved, do not delete).

### Status: ⛔ BLOCKED at Gate A on Apple Silicon (Apple **M5 Max** / macOS 26)
- ✅ arm64 native containers: PASS.
- ❌ `linux/amd64` via FEX64: **FEX SIGILLs** — does not run.

### Root cause (diagnosed live)
The HVF guest is advertised **phantom SVE it cannot execute**. `/proc/cpuinfo` in
the guest reports `sve2 sve2p1 svebf16 …`, yet a single SVE instruction (`rdvl`)
SIGILLs. FEX trusts `HWCAP_SVE`, runs SVE, and traps. This appears specific to
M5 + macOS-26 HVF (a real M1 advertises no SVE → FEX uses NEON and works).

### Fixes attempted
1. **`arm64.nosve arm64.nosme` guest cmdline** (this branch, `514f36b`,
   `app/arcbox-core/src/vm_lifecycle/mod.rs`) — ✅ verified: guest → **0 SVE
   features**. Correct hardening regardless of the amd64 decision; **keep**.
2. **apple-m1 FEX build** (boot-assets PR #23) — removed FEX's *auto-vectorized*
   SVE, but FEX-2605 also emits **explicit, unconditional SVE** (gdb:
   `ld1rd {z5.d}, p5/z` in FEX's own code) that runs and traps **even with guest
   SVE disabled**. So neither fix, nor both together, makes FEX run amd64 here.

### Decision point
Per the ABX-375 plan, a Gate-A failure → resume **ABX-374** (VZ Rosetta) for amd64
rather than duct-taping. Options:
- **A (recommended):** route `linux/amd64` runtime to VZ Rosetta (proven); keep the
  HV path + `arm64.nosve` for arm64. FEX stays optional/experimental.
- **B:** patch FEX-2605 to stop emitting unconditional SVE (or properly gate it),
  or try a newer FEX — uncertain; may be a FEX-vs-M5/HVF incompatibility.

### What's in this branch
- ABX-375 routing pivot — amd64→FEX translator, fail-closed when FEX absent, no VZ
  runtime VM (`f4387f9`, `app/arcbox-docker/src/routing.rs` + handlers).
- `assets.lock` pinned to boot-assets **v0.5.9** (static FEX at
  `/arcbox/runtime/bin/FEX`) — `c9e2e6b`.
- Reproducible gate harness — `tests/fex/validate-fex64.sh`.
- `arm64.nosve` guest fix — `514f36b`.

### Related resources
- boot-assets **PR #23** (FEX apple-m1 build + scoped SVE guard; CI-red by design —
  the guard correctly catches FEX's remaining explicit SVE):
  https://github.com/arcboxlabs/boot-assets/pull/23
- boot-assets **v0.5.9** release — the static FEX binaries that ship today.
- ABX-374 fallback (preserved): branch `feat/dual-utility-vm-routing`, PR #291.

### Reproduce the blocker
```
# in the guest (via any container):
grep -m1 Features /proc/cpuinfo | tr ' ' '\n' | grep sve   # -> sve2p1, svebf16, ...
# a one-line `rdvl` program compiled with -march=armv8-a+sve SIGILLs (exit 132).
# amd64 path:
docker --context arcbox run --rm --platform linux/amd64 alpine uname -m   # SIGILL
```


PeronGH:
## Root cause found — it's `randomize_va_space`, not SVE

The Gate-A blocker (`linux/amd64` via FEX64) is **resolved**, and the "phantom SVE / SIGILL" diagnosis in the description was a red herring — those SIGILLs were collateral from a deeper allocator failure.

**Root cause:** the guest kernel ships `CONFIG_COMPAT_BRK=y`, which forces `kernel.randomize_va_space=1` at boot (mmap/stack/vDSO randomized, brk/heap not). FEX's x86-64 allocator can't lay out its VA space under `=1` and fails non-deterministically with `Failed to map VMA region` → SIGSEGV. The SIGILLs only surfaced on runs that got far enough; once the allocator is fixed they're gone.

**How it was isolated:** OrbStack runs the *same* FEX binary cleanly — its guest has `randomize_va_space=2`; ours had `=1`. Toggling the sysctl at runtime: `1` → fails, `2` → works (`x86_64`, 10/10) on **both** a 52-bit (LPA2) and a 48-bit kernel. So VA width and page size were not the cause — the `arm64.nosve` and VA_BITS=48 detours were dead ends.

**Fix:** one line in the guest kernel config — disable `COMPAT_BRK` so it boots `randomize_va_space=2` (matching OrbStack / stock distros): arcboxlabs/kernel#7

**Result:** with a kernel built from that PR, `docker run --platform linux/amd64 alpine uname -m` → `x86_64`, deterministic, **no SVE masking and no manual sysctls**. Gate A passes.

**What changes on this branch:** keep the drift-detection fixes (recreate the default VM when kernel / cmdline / cpus / memory drift). The VMM SME-mask and the guest VA_BITS=48 kernel change explored during debugging are **unnecessary** and were reverted/dropped — `COMPAT_BRK` is the sole fix.

PR #301, opened 2026-06-12
perf(boot): cut host-side warm-boot overhead

AprilNEA:
Follow-up to #299 — reduce avoidable warm-boot overhead, make cleanup decisions explicit/tested, and move guest runtime readiness from host retry loops to a guest-driven readiness stream.

## Changes

1. **Tighten unavoidable boot discovery polling** — the remaining connect-discovery loops were paced for ~100ms-per-probe vsock; with event-driven RX a failed probe costs ~1ms, so sleep intervals dominated discovery time:
   - `wait_for_agent`: 100ms fixed → 25ms for the only unavoidable phase (discovering when the guest starts listening on the agent vsock port)
   - `wait_for_machine_ready` (`arcbox machine` VMs): probe immediately instead of sleeping 500ms before the first attempt; 50→500ms backoff under a 60s deadline

2. **Replace Docker runtime readiness polling with a guest-driven stream** — host now sends one `WatchReadiness` request and the guest emits readiness transitions on the same RPC connection:
   - new wire types: `WatchReadinessRequest` / `ReadinessEvent`
   - agent emits `AgentReady`, `RuntimeStarting`, then terminal `RuntimeReady` / `RuntimeFailed`
   - `GuestDockerBackend` uses the readiness stream as the only Docker runtime readiness path and validates the reported vsock endpoint before returning
   - no stale-agent compatibility path: daemon and guest agent are bundle-locked, so watch failures are startup protocol errors and fail fast
   - both async vsock and macOS HV blocking socketpair transports are supported without touching the tokio/kqueue reactor from the HV probe loop

3. **Use readiness events for agent handshake after connect** — once `connect_agent` succeeds, startup no longer performs a separate ping-poll loop. The readiness stream is the locked daemon↔agent contract.

4. **Type the resource-cleanup policy** — startup cleanup is now a small explicit policy instead of a compound boolean:
   - `ResourceCleanupDecision::Skip`
   - `ResourceCleanupDecision::ScanDiskImageHolders { reason }`
   - reasons distinguish `TookOverLockHolder` from `InterruptedMachineRun`

5. **Skip the disk-image holder scan only on true clean starts** — `pids_by_path` walks every process's fd table (~117ms measured) to find residual VZ XPC helpers.
   - Takeover path still scans when lock acquisition displaced a live stale daemon (`displaced_stale_daemon()`).
   - Crash/interrupted-run path also scans when persisted machine state contains `running`, covering the case where the kernel released the flock but Virtualization.framework helpers may still be unwinding and holding `docker.img`.
   - Only true clean starts skip the scan: no displaced stale daemon, no interrupted running machine state, or no persistent Docker disk image.

6. **Add startup phase timing logs** — the typed startup pipeline now records structured timing for each host-side phase:
   - `prepare_host`
   - `acquire_daemon_lease`
   - `start_control_plane`
   - `release_stale_resources`
   - `prepare_assets`
   - `boot_runtime`
   - `start_runtime_services`
   - `mark_ready`

## Review follow-up

Addressed the crash-recovery edge case called out by review: immediate flock acquisition no longer automatically means cleanup is unnecessary. The daemon now uses both signals — lock displacement and persisted interrupted machine state — before deciding whether to skip the expensive holder scan.

Also fixed the interrupted-machine signal to read the actual production persistence path (`<data_dir>/machines`), matching `MachineManager`, instead of the persistent data subdirectory.

## Tests

Added cleanup policy tests for:

- clean start with no images skips
- clean start with images skips
- takeover with images scans
- interrupted running machine with images scans
- takeover + interrupted prefers takeover reason
- interrupted signal without images skips
- disk image candidate discovery
- persisted `running` machine triggers the crash-recovery cleanup signal
- persisted `stopped` machine does not
- missing machine state does not

## Measurements (isolated daemon, warm boot, Apple Silicon)

| Phase | before | after |
|---|---|---|
| daemon start → VMM start | 524ms | **4ms**¹ |
| agent ready | 2.63–2.73s | 2.06–2.13s |
| runtime initialized | 2.98–3.10s | **2.43–2.50s** |

¹ the pre-VM phase win includes the companion `arcbox-boot` 0.5.2 change (boot-assets PR: verification cache — stops re-hashing 232MB of runtime binaries on every boot, ~410ms). That dep bump lands separately once 0.5.2 is published; this PR stands alone and accounts for ~170ms of the total.

The latest readiness-stream commit removes host retry overshoot from the Docker runtime phase; the remaining time is mostly guest kernel/runtime startup work rather than host polling cadence.

## Verification

- `cargo fmt --check`
- `cargo test -p arcbox-constants wire`
- `cargo check -p arcbox-core -p arcbox-daemon -p arcbox-agent -p arcbox-protocol -p arcbox-constants`
- `cargo clippy -p arcbox-core -p arcbox-daemon -p arcbox-agent --all-targets -- -D warnings`
- `cargo test -p arcbox-core -p arcbox-daemon -p arcbox-agent`

### DIFF
--- a/app/arcbox-core/src/agent_client.rs
+++ b/app/arcbox-core/src/agent_client.rs
@@ -12,8 +12,8 @@ use arcbox_protocol::agent::{
     KubernetesKubeconfigRequest, KubernetesKubeconfigResponse, KubernetesStartRequest,
     KubernetesStartResponse, KubernetesStatusRequest, KubernetesStatusResponse,
     KubernetesStopRequest, KubernetesStopResponse, MmapReadFileRequest, MmapReadFileResponse,
-    PingRequest, PingResponse, RuntimeEnsureRequest, RuntimeEnsureResponse, RuntimeStatusRequest,
-    RuntimeStatusResponse, SystemInfo,
+    PingRequest, PingResponse, ReadinessEvent, RuntimeEnsureRequest, RuntimeEnsureResponse,
+    RuntimeStatusRequest, RuntimeStatusResponse, SystemInfo, WatchReadinessRequest,
 };
 use arcbox_protocol::sandbox_v1::{
     CheckpointRequest, CheckpointResponse, CreateSandboxRequest, CreateSandboxResponse,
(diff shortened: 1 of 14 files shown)


AprilNEA:
Rebased this PR onto latest `master` and addressed the crash-recovery edge case called out in review.

Change added on top:

- clean starts still skip the expensive `pids_by_path` holder scan
- takeovers still scan via `displaced_stale_daemon()`
- crash/interrupted-run starts now also scan when persisted machine state contains `running`, covering the case where the kernel released the flock but VZ XPC helpers may still be unwinding and holding `docker.img`

Added daemon startup tests for:

- persisted `running` machine triggers the cleanup signal
- persisted `stopped` machine does not
- missing machine state does not

Validated:

- `cargo fmt --check`
- `cargo test -p arcbox-daemon startup::tests`
- `cargo check -p arcbox-core -p arcbox-daemon`
- `cargo clippy -p arcbox-core -p arcbox-daemon --all-targets -- -D warnings`
- `cargo test -p arcbox-core -p arcbox-daemon`

AprilNEA:
Pushed a more structural warm-boot governance pass on top of the rebase.

What changed beyond the earlier fix:

- replaced the ad-hoc cleanup boolean with a typed `ResourceCleanupDecision`
- distinguished cleanup reasons: `TookOverLockHolder` vs `InterruptedMachineRun`
- fixed interrupted-machine detection to read the real production path: `<data_dir>/machines`
- added cross-platform cleanup policy matrix tests
- added structured startup phase timing logs (`startup_phase`, `elapsed_ms`) across the typed pipeline

What I intentionally did not fake in this PR:

- guest readiness events: needs protocol + guest agent watch support
- startup journal: needs durable state/corruption/fsync design
- VM snapshot / keep-warm: changes lifecycle and consistency contracts
- hard perf gates: need measured baselines first

Validated:

- `cargo fmt --check`
- `cargo test -p arcbox-daemon startup::resource_cleanup`
- `cargo check -p arcbox-core -p arcbox-daemon`
- `cargo clippy -p arcbox-core -p arcbox-daemon --all-targets -- -D warnings`
- `cargo test -p arcbox-core -p arcbox-daemon`

AprilNEA:
Pushed the broader warm-boot follow-up in `457711d`.

What changed beyond the earlier polling cleanup:

- Added a real guest readiness stream to the agent protocol: `WatchReadinessRequest` → `ReadinessEvent`.
- `GuestDockerBackend` now uses one long-lived readiness watch as the primary Docker runtime readiness path instead of host-side ensure/status retry polling.
- The guest emits `AgentReady`, `RuntimeStarting`, then terminal `RuntimeReady` / `RuntimeFailed` events and keeps observing runtime readiness until the caller's timeout.
- Startup uses the same readiness event path after `connect_agent` succeeds, so the old ping loop is only a compatibility fallback for stale agents.
- Supported both async vsock and the macOS HV blocking socketpair transport without reintroducing tokio/kqueue reactor stalls.

Validation rerun:

- `cargo fmt --check`
- `cargo test -p arcbox-constants wire`
- `cargo check -p arcbox-core -p arcbox-daemon -p arcbox-agent -p arcbox-protocol -p arcbox-constants`
- `cargo clippy -p arcbox-core -p arcbox-daemon -p arcbox-agent --all-targets -- -D warnings`
- `cargo test -p arcbox-core -p arcbox-daemon -p arcbox-agent`

AprilNEA:
Follow-up pushed in `9205116`: removed the stale-agent compatibility paths.

Since daemon and guest agent are bundle-locked, readiness watch is now treated as the protocol contract:

- `GuestDockerBackend` no longer falls back to host-side `ensure_runtime` / `get_runtime_status` polling.
- `wait_for_agent` no longer falls back to ping after `connect_agent` succeeds.
- A readiness-watch failure now fails startup directly, which should expose protocol drift instead of hiding it behind a slower legacy path.

Re-verified:

- `cargo fmt --check`
- `cargo check -p arcbox-core -p arcbox-daemon -p arcbox-agent`
- `cargo clippy -p arcbox-core -p arcbox-daemon -p arcbox-agent --all-targets -- -D warnings`
- `cargo test -p arcbox-core -p arcbox-daemon -p arcbox-agent`

PR #302, opened 2026-06-12
fix(docker): cache guest HTTP readiness in proxy

AprilNEA:
Follow-up to #299/#301 on the Docker request path.

## Problem

Every proxied Docker API request re-ran the slow guest readiness path: VM lifecycle readiness, guest-agent `ensure_runtime`, status fallback, and a throwaway endpoint probe. That made warm Docker requests pay agent/vsock round trips even after guest `dockerd` was already known to be serving traffic.

The earlier approach cached readiness inside `arcbox-core`, but that put Docker HTTP semantics in the wrong layer and could only prove transport/connectivity. The proxy layer is the component that can prove guest `dockerd` accepted and answered a real Docker request.

## Change

Move Docker HTTP readiness ownership into `arcbox-docker`:

- `arcbox-core` now owns only the slow readiness path: VM lifecycle, guest-agent runtime readiness, and reported vsock endpoint validation.
- `arcbox-docker::ProxyState` owns HTTP verification and invalidation.
- Guest dockerd readiness is tracked per `UtilityVmRole` with an explicit state machine:
  - `Unverified`
  - `Verifying`
  - `Verified`
- First request for a role runs the slow readiness path, then verifies guest dockerd with a real `GET /_ping` through the pooled proxy client.
- Concurrent callers that observe `Verifying` wait for the state change instead of starting duplicate probes.
- Forwarding, upload, or upgrade transport failures invalidate that role's readiness so the next request re-runs slow readiness + HTTP verification.

## Why this shape

This keeps the ownership boundary clean:

- `arcbox-core`: VM/agent/runtime orchestration and endpoint shape.
- `arcbox-docker`: Docker HTTP readiness, readiness caching, and transport-failure invalidation.

It also avoids the previous connect-only readiness ambiguity: `Verified` now means guest `dockerd` answered an actual Docker HTTP request.

## Tests

Added readiness state tests for:

- `Unverified -> Verified` on successful prepare + `_ping`
- cache hit while already `Verified`
- failure rollback to `Unverified`
- invalidation from `Verified` to `Unverified`
- per-role isolation
- concurrent verification serialization

## Verification

- `cargo fmt --check`
- `cargo check -p arcbox-core -p arcbox-docker`
- `cargo clippy -p arcbox-core -p arcbox-docker --all-targets -- -D warnings`
- `cargo test -p arcbox-core -p arcbox-docker`

## Docs

Updated `app/arcbox-docker/README.md` with the proxy readiness ownership model, state machine, and focused test command.

### DIFF
--- a/app/arcbox-core/src/container_backend/guest_docker.rs
+++ b/app/arcbox-core/src/container_backend/guest_docker.rs
@@ -6,6 +6,7 @@ use crate::vm_lifecycle::VmLifecycleManager;
 use async_trait::async_trait;
 use std::os::fd::FromRawFd;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::{Duration, Instant};
 
 /// Guest Docker backend (dockerd/containerd/runc inside VM).
@@ -14,6 +15,9 @@ pub struct GuestDockerBackend {
     machine_manager: Arc<MachineManager>,
     machine_name: &'static str,
     config: ContainerRuntimeConfig,
+    /// Set once the guest endpoint has been verified; collapses the
+    /// per-request readiness check to a single vsock connect probe.
+    endpoint_verified: AtomicBool,
 }
 
 impl GuestDockerBackend {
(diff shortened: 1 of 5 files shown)


AprilNEA:
Handled the review feedback:

- The cached readiness probe now performs a real Docker `_ping` request over the vsock connection instead of treating connect-only as proof of readiness. This covers HV-backed VMs where connect only allocates/enqueues the socketpair and the guest RESPONSE/RST is observed through later I/O.
- The probe runs in `spawn_blocking` and is bounded by a short timeout so the warm async request path does not block a Tokio worker on connect/read/write.
- Added an explicit comment explaining why `Ordering::Relaxed` is intentional for `endpoint_verified`.
- Reused the same `_ping` probe for the slow-path final endpoint validation before setting the cache bit.
- Added parser tests for expected/non-200 Docker `_ping` responses.

Local validation passed: `cargo fmt --check`, `cargo check -p arcbox-core`, `cargo clippy -p arcbox-core --all-targets -- -D warnings`, `cargo test -p arcbox-core`.

AprilNEA:
Updated this PR with the structural readiness refactor:

- moved Docker HTTP verification out of arcbox-core and into the docker proxy state
- core now only drives VM/agent/runtime readiness and validates the reported vsock endpoint
- proxy caches per-role guest dockerd HTTP readiness after a real GET /_ping
- proxy invalidates the cached readiness on forwarding/upload/upgrade transport failures so the next request re-runs slow runtime readiness + HTTP verification
- added readiness cache/invalidation tests

Validated:
- cargo fmt --check
- cargo check -p arcbox-core -p arcbox-docker
- cargo clippy -p arcbox-core -p arcbox-docker --all-targets -- -D warnings
- cargo test -p arcbox-core -p arcbox-docker

AprilNEA:
Follow-up to make the readiness model explicit:

- replaced the proxy readiness AtomicBool/gate with a typed EndpointReadinessState: Unverified -> Verifying -> Verified
- concurrent callers now wait on the Verifying state via Notify instead of starting duplicate probes
- added state-transition tests for success, cache-hit, failure rollback, invalidation, per-role isolation, and concurrent verification serialization

Revalidated:
- cargo fmt --check
- cargo clippy -p arcbox-core -p arcbox-docker --all-targets -- -D warnings
- cargo test -p arcbox-core -p arcbox-docker

PR #303, opened 2026-06-12
fix(vmnet): match vmpktdesc ABI so vmnet_read/vmnet_write stop failing EINVAL

AprilNEA:
## Problem

After #298 shipped vmnet in default features (v0.4.6), `*.arcbox.local` domains still did not appear in the desktop UI. Debug logs showed the bridge NIC relay failing every single read with:

```
vmnet read error: invalid argument
```

## Root cause

`VmnetPacket` in `arcbox-vmnet/src/ffi.rs` did not match `struct vmpktdesc` from `<vmnet/vmnet.h>`:

```c
struct vmpktdesc {
    size_t        vm_pkt_size;   // 8 bytes, FIRST field
    struct iovec *vm_pkt_iov;
    uint32_t      vm_pkt_iovcnt;
    uint32_t      vm_flags;
};
```

The Rust struct declared `vm_pkt_size` as a `u32` in third position. Every field landed at the wrong offset: vmnet read the iov pointer as the packet size, garbage as the iov pointer, and 0 as the iov count — so **both** `vmnet_read` and `vmnet_write` returned `VMNET_INVALID_ARGUMENT` for every packet. The NIC2 data path was dead in both directions:

- guest eth1 never completed DHCP → no IP
- kernel FDB never learned the guest MAC → `resolve_bridge_by_mac` returned None
- container route `172.16/12 → bridge100` was never installed
- `route_installed` stayed false → desktop UI showed `localhost` instead of `<name>.arcbox.local`

## Fix

- Reorder `VmnetPacket` to match the C layout; `vm_pkt_size` is now `usize` (= `size_t`)
- Set `vm_pkt_size` to the buffer capacity on read (vmnet_read's documented in/out contract) instead of 0
- Add an `offset_of!`-based ABI test pinning size 24 / offsets 0, 8, 16, 20

## Verification (macOS 26.5, end-to-end)

With this fix, on a cold boot:

- the EINVAL flood is gone (0 occurrences vs ~55k before)
- guest eth1 acquires `192.168.64.2/24` from vmnet's DHCP
- host ARP resolves the guest's vmnet MAC on `bridge100`; host↔guest ICMP flows
- FDB lookup resolves `bridge100`; helper logs `route ensured subnet=172.16.0.0/12 bridge=bridge100`
- `arcbox-postgres-1.arcbox.local` resolves to its container IP via the resolver at 127.0.0.1:5553

## Follow-ups (not in this PR)

1. **Route conflict**: on LANs that themselves route `172.16/12` (mine pushes it to the LAN gateway), configd re-asserts the DHCP route and silently replaces the helper's bridge100 route minutes after install. The route reconciler needs to watch the routing socket and re-assert, or use a less conflict-prone strategy.
2. `VmnetReturnT` is missing `VMNET_NOT_AUTHORIZED` (1010); an unknown discriminant in the `repr(C)` enum returned by FFI is UB.
3. `relay.rs` guest→vmnet loop: the socketpair fd is never set `O_NONBLOCK` and the `AsyncFd` guard never calls `clear_ready()` after successful reads — works, but degenerates into blocking reads on a runtime worker.
4. Pre-existing cold-boot flakiness ("timeout waiting for agent"): the guest agent intermittently stalls in early init; today the app's own daemon needed 3+ launchd restarts before a boot succeeded (12:48–12:52 in logs). Unrelated to vmnet — tracked separately.
5. vmnet ignores `with_mac` (assigns a random MAC per boot), so `recovery.rs`'s FDB lookup by the *derived* MAC can never match — live path already uses the vmnet-reported MAC.

### DIFF
--- a/virt/arcbox-vmnet/src/ffi.rs
+++ b/virt/arcbox-vmnet/src/ffi.rs
@@ -47,6 +47,8 @@ pub enum VmnetReturnT {
     TooManyPackets = 1008,
     /// Sharing service busy.
     SharingServiceBusy = 1009,
+    /// The process is not authorized to use vmnet.
+    NotAuthorized = 1010,
 }
 
 impl VmnetReturnT {
@@ -70,6 +72,7 @@ impl VmnetReturnT {
             Self::BufferExhausted => "buffer exhausted",
             Self::TooManyPackets => "too many packets",
             Self::SharingServiceBusy => "sharing service busy",
+            Self::NotAuthorized => "not authorized",
         }
     }
 }
@@ -549,8 +552,16 @@ mod tests {
     fn test_vmnet_return_messages() {
         assert_eq!(VmnetReturnT::Success.message(), "success");
         assert_eq!(VmnetReturnT::Failure.message(), "operation failed");
+        assert_eq!(VmnetReturnT::NotAuthorized.message(), "not authorized");
         assert!(VmnetReturnT::Success.is_success());
         assert!(!VmnetReturnT::Failure.is_success());
+    }
+
+    #[test]
+    fn test_vmnet_return_values() {
+        assert_eq!(VmnetReturnT::Success as i32, 1000);
+        assert_eq!(VmnetReturnT::SharingServiceBusy as i32, 1009);
+        assert_eq!(VmnetReturnT::NotAuthorized as i32, 1010);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


AprilNEA:
Handled the remaining review feedback:

- Added `VmnetReturnT::NotAuthorized = 1010` with message/test coverage, closing the missing vmnet status UB gap noted by Greptile.
- Gated the hardcoded `VmnetPacket` ABI layout test with `#[cfg(target_pointer_width = "64")]` so the 64-bit macOS assumption is explicit.
- Resolved the inline review thread.

Local validation passed: `cargo fmt --check`, `cargo check -p arcbox-vmnet`, `cargo clippy -p arcbox-vmnet --all-targets -- -D warnings`, `cargo test -p arcbox-vmnet`.

PR #306, opened 2026-06-15
fix(daemon): refresh staged binaries when the bundle's copy changes

AprilNEA:
## Problem
`seed_from_bundle` and `ensure_agent_binary` copied the agent, runtime binaries, and boot assets into `~/.arcbox/` **only when the destination was absent** (`!dst.exists()`). Because the agent (`bin/arcbox-agent`) and runtime tree (`runtime/`) are not versioned by path, once staged they were **never refreshed** — so after an app update the daemon kept running stale staged binaries.

This bit us in the field: a stale guest agent (from a previous version) persisted across an update and failed to come up against the newer kernel/rootfs, leaving the daemon crash-looping on `timeout waiting for agent`. The bundled runtime binaries (dockerd/containerd/runc/FEX/…) had the same latent staleness.

(Boot assets dodged it only because they live under a version-keyed path `boot/{version}/`.)

## Fix
Add `copy_if_changed(src, dst)`: copies when `dst` is missing or differs from `src` by **size or mtime**, and mirrors the source mtime onto the destination so later startups detect an up-to-date copy with a cheap `stat` — no content read, which matters because seeding runs on every daemon start and the runtime tree is hundreds of MB. `std::fs::copy` carries over the source permission bits (incl. the exec bit). Route agent, runtime, and boot seeding through it; this also removes the now-redundant manual exec-bit handling.

## Verification
- New unit tests (`copy_if_changed`): copies-when-missing then skips-when-identical, recopies-on-source-change, preserves-exec-bit — all green.
- `cargo clippy -p arcbox-daemon --all-targets` clean; `cargo fmt` clean; `cargo test -p arcbox-daemon` 9 passed.

Context: this is the latent bug behind the agent incident also addressed operationally alongside arcbox-desktop#234 / arcbox#304.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -336,12 +336,14 @@ dependencies = [
  "bytes",
  "clap",
  "dirs",
+ "filetime",
  "http-body-util",
  "libc",
  "libproc",
  "macos-resolver",
  "sentry",
  "serde_json",
+ "tempfile",
  "tokio",
  "tokio-stream",
  "tokio-util",
@@ -1664,13 +1666,12 @@ dependencies = [
 
 [[package]]
 name = "filetime"
-version = "0.2.27"
+version = "0.2.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f98844151eee8917efc50bd9e8318cb963ae8b297431495d3f758616ea5c57db"
+checksum = "5c287a33c7f0a620c38e641e7f60827713987b3c0f26e8ddc9462cc69cf75759"
 dependencies = [
  "cfg-if",
  "libc",
- "libredox",
 ]
 
 [[package]]
@@ -2493,10 +2494,7 @@ version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ddbf48fd451246b1f8c2610bd3b4ac0cc6e149d89832867093ab69a17194f08"
 dependencies = [
- "bitflags 2.11.0",
  "libc",
- "plain",
- "redox_syscall 0.7.3",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


AprilNEA:
Updated this PR with a safer agent staging flow:

- bundle-provided arcbox-agent is now authoritative and startup fails if it cannot be synced
- boot-cache arcbox-agent is only used as fallback when the bundle has no agent
- fallback staging now happens after boot assets are ensured, so fresh non-bundle installs can stage the downloaded agent
- added regression coverage for both fallback install and preventing boot cache from overwriting a bundle agent

Local validation passed: cargo fmt --check, cargo check -p arcbox-daemon, cargo clippy -p arcbox-daemon --all-targets -- -D warnings, cargo test -p arcbox-daemon.

AprilNEA:
Pushed the structural startup refactor on top of the agent refresh fix.

What changed:
- main.rs now only handles process entry/logging and delegates lifecycle to startup::Startup.
- Added a typed startup pipeline: host prepared -> daemon leased -> control plane started -> stale resources released -> assets prepared -> runtime booted -> runtime services started -> ready.
- Asset reconciliation is now an explicit prepare_assets stage before Runtime init, with PreparedAssets/PreparedAgent recording whether the staged agent came from the bundle, boot cache, or was unavailable.
- Kept bundle agent precedence from the previous commit: bundle agent remains authoritative; boot cache is fallback only.
- Rebasing onto current master exposed a clippy ordering issue in recovery.rs, so I moved its test module to the file end without behavior changes.

Local validation after rebase passed: cargo fmt --check, cargo check -p arcbox-daemon, cargo clippy -p arcbox-daemon --all-targets -- -D warnings, cargo test -p arcbox-daemon.

PR is mergeable again; GitHub CI has restarted.

PR #309, opened 2026-06-16
fix(bundle):add diasble apple lib validation

lucas77778:
issue #308

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,6 +160,11 @@ make run-daemon
 | `codesign` reports `errSecInternalComponent` | Private key missing (only cert imported) | Re-export `.p12` with both cert **and** private key selected |
 | Daemon starts but `--options runtime` was omitted | Restricted entitlements silently ignored | Always include `--options runtime` in the codesign command |
 
+> [!NOTE]
+> For deeper diagnostics — chain-to-self-signed-root errors, manual trust
+> overrides, and suppressing the per-sign keychain password prompt — see
+> [`docs/code-signing-troubleshooting.md`](docs/code-signing-troubleshooting.md).
+
 > [!WARNING]
 > The `--options runtime` flag enables Hardened Runtime, which is **required**
 > for restricted entitlements to take effect. Without it, macOS accepts the
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,10 @@
 # ArcBox Development Makefile
 
 PROFILE ?= debug
-ENTITLEMENTS := bundle/arcbox.entitlements
+# Local dev signing uses the dev entitlements (disables library validation
+# for nix-store libiconv). The release path (.github/workflows/release.yml)
+# signs with bundle/arcbox.entitlements, which keeps validation enabled.
+ENTITLEMENTS := bundle/arcbox.dev.entitlements
 AGENT_TARGET := aarch64-unknown-linux-musl
 
 # Signing identity: auto-detect "Developer ID Application: ArcBox, Inc."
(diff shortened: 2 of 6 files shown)


AprilNEA:
I suspect the root cause here may not actually require a separate dev-only entitlements path. It looks possible that the explicit Darwin `libiconv` dependency is causing `arcbox-daemon` to link against the real dylib from the Nix store.

Could you help validate a smaller fix first?

1. Check which `libiconv` the current daemon binary is actually linked against:
```bash
otool -L target/debug/arcbox-daemon | grep -i iconv
```

2. If that shows `/nix/store/.../libiconv...`
   - that is likely the direct trigger for Hardened Runtime library validation
   - in that case, could you try removing `pkgs.libiconv` from `devenv.nix`, rebuilding, and checking `otool -L` again?

3. If the build still succeeds after removing it, and the linked path becomes `/usr/lib/libiconv.2.dylib`
   - then a good portion of the current dev-entitlements split may be working around something we can eliminate at link time instead
   - that would let us shrink the fix quite a bit

I’m not opposed to the current direction, but I’d like to confirm whether we’re fixing the actual root cause or adding a runtime exemption for a problem introduced by explicitly linking the Nix `libiconv`. If removing `pkgs.libiconv` turns out not to be viable, the current split-entitlements approach becomes much easier to justify.

lucas77778:
> I suspect the root cause here may not actually require a separate dev-only entitlements path. It looks possible that the explicit Darwin `libiconv` dependency is causing `arcbox-daemon` to link against the real dylib from the Nix store.
> 
> Could you help validate a smaller fix first?
> 
> 1. Check which `libiconv` the current daemon binary is actually linked against:
> 
> ```shell
> otool -L target/debug/arcbox-daemon | grep -i iconv
> ```
> 
> 2. If that shows `/nix/store/.../libiconv...`
>    
>    * that is likely the direct trigger for Hardened Runtime library validation
>    * in that case, could you try removing `pkgs.libiconv` from `devenv.nix`, rebuilding, and checking `otool -L` again?
> 3. If the build still succeeds after removing it, and the linked path becomes `/usr/lib/libiconv.2.dylib`
>    
>    * then a good portion of the current dev-entitlements split may be working around something we can eliminate at link time instead
>    * that would let us shrink the fix quite a bit
> 
> I’m not opposed to the current direction, but I’d like to confirm whether we’re fixing the actual root cause or adding a runtime exemption for a problem introduced by explicitly linking the Nix `libiconv`. If removing `pkgs.libiconv` turns out not to be viable, the current split-entitlements approach becomes much easier to justify.

Root cause: pkgs.libiconv isn't the source of the linkage. The path is injected by the nix darwin toolchain via NIX_LDFLAGS, independent of the packages list:
$ echo "$NIX_LDFLAGS" | tr ' ' '\n' | grep iconv
-L/nix/store/xvmhkpvfvmy4sfdkqwg9inq3qkpnx81b-libiconv-109.100.2/lib
This -L is propagated by the nix darwin stdenv / Rust toolchain / apple-sdk_26 (it sits alongside the -L flags for libcxx, compiler-rt, the rust toolchain, etc.), so the linker resolves libiconv.2.dylib from the nix store ahead of /usr/lib regardless of whether pkgs.libiconv is listed.

So the link-time fix you proposed doesn't change the outcome — eliminating it would mean overriding the toolchain's NIX_LDFLAGS / forcing /usr/lib ahead of the nix store, which is more invasive and brittle. Given that, the split-entitlements approach looks like the more defensible direction

AprilNEA:
dev entitlement is dangerous, we need to set up the gate later.

AprilNEA (approved):

PR #310, opened 2026-06-16
feat(fleet): cross-platform runner agent

PeronGH:
## What

Adds a standalone, cross-platform Fleet runner agent that enrolls a host with the Fleet gateway and runs GitHub Actions jobs directly (no isolation in v1 — the job is handed to the pre-installed runner via `run.sh --jitconfig`).

New `fleet/` layer (shares no code with the macOS virt workspace):
- **`arcbox-fleet-proto`** — tonic client stubs generated from the public BSR contract `buf.build/arcboxlabs/fleet` (vendored proto; `make fleet-proto-sync` to refresh). Client-only.
- **`arcbox-fleet-agent`** — env-configured binary (`ARCBOX_FLEET_*`):
  - `enroll --token` → exchanges the token, persists a `0600` credential.
  - `run` → durable `Attach` stream (heartbeat out, dispatch in, backoff reconnect); on `ProvisionRunner` spawns the runner and reports `RunnerAccepted/Started/Finished/Failed`.
  - `ProvisionRunner.image/cpus/mem_mib` are no-ops in v1.

Durability:
- The runner supervisor and its outbound event queue outlive any single connection, so a job dispatched before a reconnect keeps running and its terminal event reaches the next live stream instead of a dropped connection's dead channel.
- `ProvisionRunner` handling is idempotent on `job_id`: the gateway's dispatch is at-least-once "pure transport", so a redelivered order for an in-flight job is a no-op rather than a second runner process.

## Verification

Full E2E on a Linux box against the real control plane (`arcbox-api` + `arcbox-worker` + `arcbox-fleet-gateway` + Restate + Postgres/Redis), with the in-repo `github-mock`/`gh-webhook` harness standing in for GitHub:

- ✅ **Enroll** → `0600` credential persisted; gateway logs `machine enrolled`.
- ✅ **Attach + heartbeat** → machine `online`, `last_seen` set, capacity pool recorded.
- ✅ **Dispatch → run → report** → `gh-webhook queued` drives the real webhook→worker→provision→dispatch chain; the agent runs the JIT runner and reports `RunnerAccepted/Started/Finished`; `runner_jobs` reaches `completed`.
- ✅ **Reconnect durability** → bouncing the gateway mid-run delivers the in-flight job's `RunnerFinished` on the new stream (gateway logs `runner finished` after the reattach; no dropped event).
- ✅ **Duplicate dispatch** → a redelivered `ProvisionRunner` for an in-flight job is ignored (`duplicate provision ignored`), single runner spawn.
- ✅ `cargo fmt` / `clippy` / `test` clean per-crate.

## Notes
- Full client workspace doesn't cross-compile on Linux (macOS crates); build in isolation with `cargo build -p arcbox-fleet-agent`. fmt/clippy/test were run per-crate.

### DIFF
--- a/fleet/arcbox-fleet-agent/src/attach.rs
+++ b/fleet/arcbox-fleet-agent/src/attach.rs
@@ -13,6 +13,7 @@ use arcbox_fleet_proto::v1::fleet_gateway_service_client::FleetGatewayServiceCli
 use arcbox_fleet_proto::v1::{AttachRequest, Heartbeat, attach_request, attach_response};
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
+use tokio_util::sync::CancellationToken;
 use tonic::Request;
 use tonic::metadata::MetadataValue;
 use tracing::{info, warn};
@@ -27,13 +28,24 @@ const INITIAL_BACKOFF: Duration = Duration::from_secs(1);
 const MAX_BACKOFF: Duration = Duration::from_secs(60);
 const OUTBOUND_CAPACITY: usize = 64;
 const MACHINE_TOKEN_HEADER: &str = "x-arcbox-machine-token";
+/// How long to wait for runner process groups to be killed and reaped on
+/// shutdown before giving up. Reaping a SIGKILLed group is near-instant, so this
+/// is a generous ceiling rather than an expected wait.
+const SHUTDOWN_GRACE: Duration = Duration::from_secs(15);
 
-/// Connect and serve the attach stream forever, reconnecting on any failure.
+/// Connect and serve the attach stream, reconnecting on any failure until
+/// `shutdown` fires, then stop runners cleanly.
 ///
 /// The supervisor and the egress queue carrying runner lifecycle events are
 /// built once and reused across reconnects, so in-flight jobs survive a dropped
-/// connection and their terminal events reach the next live stream.
-pub async fn run(config: AgentConfig, credential: Credential) -> Result<()> {
+/// connection and their terminal events reach the next live stream. On shutdown
+/// the loop exits and hands off to [`RunnerSupervisor::shutdown`], which kills
+/// and reaps any in-flight runner process groups.
+pub async fn run(
+    config: AgentConfig,
+    credential: Credential,
+    shutdown: CancellationToken,
+) -> Result<()> {
     let runner_dir = config.require_runner_dir()?.to_path_buf();
 
     // Runner lifecycle events flow through this queue, which outlives any single
@@ -48,25 +60,39 @@ pub async fn run(config: AgentConfig, credential: Credential) -> Result<()> {
     // event is never lost to a closed stream.
     let mut pending: Option<AttachRequest> = None;
 
-    loop {
-        match connect_and_serve(
+    while !shutdown.is_cancelled() {
+        let outcome = connect_and_serve(
             &config,
             &credential,
             &supervisor,
             &mut egress_rx,
             &mut pending,
             &mut backoff,
+            &shutdown,
         )
-        .await
-        {
+        .await;
+        // A shutdown during the connection is a clean exit, not a failure to log
+        // or back off from.
+        if shutdown.is_cancelled() {
+            break;
+        }
+        match outcome {
             Ok(()) => info!("attach stream closed by gateway; reconnecting"),
             Err(e) => {
                 warn!(error = %e, backoff_secs = backoff.as_secs(), "attach failed; retrying");
             }
         }
-        tokio::time::sleep(backoff).await;
+        tokio::select! {
+            biased;
+            () = shutdown.cancelled() => break,
+            () = tokio::time::sleep(backoff) => {}
+        }
         backoff = (backoff * 2).min(MAX_BACKOFF);
     }
+
+    info!("shutdown signal received; stopping runners");
+    supervisor.shutdown(SHUTDOWN_GRACE).await;
+    Ok(())
 }
 
 /// One connect + stream lifetime. Returns `Ok(())` on a clean close. Resets
@@ -85,6 +111,7 @@ async fn connect_and_serve(
     egress_rx: &mut mpsc::Receiver<AttachRequest>,
     pending: &mut Option<AttachRequest>,
     backoff: &mut Duration,
+    shutdown: &CancellationToken,
 ) -> Result<()> {
     let channel = config
         .endpoint()?
@@ -122,6 +149,7 @@ async fn connect_and_serve(
 
     let outcome = loop {
         tokio::select! {
+            () = shutdown.cancelled() => break Ok(()),
             event = egress_rx.recv() => match event {
                 Some(msg) => {
                     if let Err(err) = req_tx.send(msg).await {
--- a/fleet/arcbox-fleet-agent/src/attach.rs
+++ b/fleet/arcbox-fleet-agent/src/attach.rs
@@ -1,0 +1,126 @@
+//! The durable outbound `Attach` stream: heartbeats out, dispatch in, with
+//! exponential-backoff reconnect.
+
+use std::path::PathBuf;
+use std::time::Duration;
+
+use anyhow::{Context, Result};
+use arcbox_fleet_proto::v1::fleet_gateway_service_client::FleetGatewayServiceClient;
+use arcbox_fleet_proto::v1::{AttachRequest, Heartbeat, attach_request, attach_response};
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::Request;
+use tonic::metadata::MetadataValue;
+use tracing::{info, warn};
+
+use crate::config::AgentConfig;
+use crate::credentials::Credential;
+use crate::host;
+use crate::runner::RunnerSupervisor;
+
+const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(15);
+const INITIAL_BACKOFF: Duration = Duration::from_secs(1);
+const MAX_BACKOFF: Duration = Duration::from_secs(60);
+const OUTBOUND_CAPACITY: usize = 64;
+const MACHINE_TOKEN_HEADER: &str = "x-arcbox-machine-token";
+
+/// Connect and serve the attach stream forever, reconnecting on any failure.
+pub async fn run(config: AgentConfig, credential: Credential) -> Result<()> {
+    let runner_dir = config.require_runner_dir()?.to_path_buf();
+    let mut backoff = INITIAL_BACKOFF;
+
+    loop {
+        match attach_once(&config, &credential, runner_dir.clone()).await {
+            Ok(()) => {
+                info!("attach stream closed by gateway; reconnecting");
+                backoff = INITIAL_BACKOFF;
+            }
+            Err(e) => {
+                warn!(error = %e, backoff_secs = backoff.as_secs(), "attach failed; retrying")
+            }
+        }
+        tokio::time::sleep(backoff).await;
+        backoff = (backoff * 2).min(MAX_BACKOFF);
+    }
+}
+
+/// One connect + stream lifetime. Returns `Ok(())` on a clean close.
+async fn attach_once(
+    config: &AgentConfig,
+    credential: &Credential,
+    runner_dir: PathBuf,
+) -> Result<()> {
+    let channel = config
+        .endpoint()?
+        .connect()
+        .await
+        .with_context(|| format!("connecting to {}", config.gateway))?;
+    let mut client = FleetGatewayServiceClient::new(channel);
+
+    let (outbound, rx) = mpsc::channel::<AttachRequest>(OUTBOUND_CAPACITY);
+    let supervisor = RunnerSupervisor::new(outbound.clone(), runner_dir, config.max_concurrent);
+    let heartbeat = spawn_heartbeat(outbound, config.max_concurrent);
+
+    let mut request = Request::new(ReceiverStream::new(rx));
+    let token: MetadataValue<_> = credential
+        .machine_token
+        .parse()
+        .context("machine token is not a valid metadata value")?;
+    request.metadata_mut().insert(MACHINE_TOKEN_HEADER, token);
+
+    let mut inbound = client
+        .attach(request)
+        .await
+        .context("Attach RPC failed")?
+        .into_inner();
+    info!("attached to gateway");
+
+    let outcome = loop {
+        match inbound.message().await {
+            Ok(Some(message)) => dispatch(&supervisor, message.msg).await,
+            Ok(None) => break Ok(()),
+            Err(status) => break Err(anyhow::Error::from(status)),
+        }
+    };
+
+    heartbeat.abort();
+    outcome
+}
+
+/// Route one inbound `AttachResponse` to the supervisor.
+async fn dispatch(supervisor: &RunnerSupervisor, msg: Option<attach_response::Msg>) {
+    match msg {
+        Some(attach_response::Msg::ProvisionRunner(order)) => {
+            supervisor.handle_provision(order).await;
+        }
+        Some(attach_response::Msg::Cancel(cancel)) => supervisor.handle_cancel(&cancel.job_id),
+        Some(attach_response::Msg::Drain(_)) => supervisor.handle_drain(),
+        None => {}
+    }
+}
+
+/// Periodically push a declarative capacity heartbeat until the channel closes.
+fn spawn_heartbeat(
+    outbound: mpsc::Sender<AttachRequest>,
+    max_concurrent: usize,
+) -> tokio::task::JoinHandle<()> {
+    tokio::spawn(async move {
+        // `interval` fires its first tick immediately, so the gateway sees us
+        // online without waiting a full period.
+        let mut ticker = tokio::time::interval(HEARTBEAT_INTERVAL);
+        loop {
+            ticker.tick().await;
+            let msg = attach_request::Msg::Heartbeat(Heartbeat {
+                capacities: host::capacities(max_concurrent),
+                host_info_json: host::host_info_json(),
+            });
+            if outbound
+                .send(AttachRequest { msg: Some(msg) })
+                .await
+                .is_err()
+            {
+                break;
+            }
+        }
+    })
+}
--- a/fleet/arcbox-fleet-agent/src/attach.rs
+++ b/fleet/arcbox-fleet-agent/src/attach.rs
@@ -30,13 +30,10 @@ pub async fn run(config: AgentConfig, credential: Credential) -> Result<()> {
     let mut backoff = INITIAL_BACKOFF;
 
     loop {
-        match attach_once(&config, &credential, runner_dir.clone()).await {
-            Ok(()) => {
-                info!("attach stream closed by gateway; reconnecting");
-                backoff = INITIAL_BACKOFF;
-            }
+        match attach_once(&config, &credential, runner_dir.clone(), &mut backoff).await {
+            Ok(()) => info!("attach stream closed by gateway; reconnecting"),
             Err(e) => {
-                warn!(error = %e, backoff_secs = backoff.as_secs(), "attach failed; retrying")
+                warn!(error = %e, backoff_secs = backoff.as_secs(), "attach failed; retrying");
             }
         }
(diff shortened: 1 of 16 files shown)


fleet/arcbox-fleet-agent/src/attach.rs
```diff
@@ -0,0 +1,126 @@
...
+/// Connect and serve the attach stream forever, reconnecting on any failure.
+pub async fn run(config: AgentConfig, credential: Credential) -> Result<()> {
+    let runner_dir = config.require_runner_dir()?.to_path_buf();
+    let mut backoff = INITIAL_BACKOFF;
+
+    loop {
+        match attach_once(&config, &credential, runner_dir.clone()).await {
+            Ok(()) => {
+                info!("attach stream closed by gateway; reconnecting");
+                backoff = INITIAL_BACKOFF;
```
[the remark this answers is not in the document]
PeronGH: Fixed. `attach_once` now takes `&mut backoff` and resets it to `INITIAL_BACKOFF` the moment the `Attach` RPC succeeds, so a connection that establishes and later drops with an error reconnects promptly instead of inheriting the escalated delay. Connect failures (never established) still back off as before.

fleet/arcbox-fleet-agent/src/config.rs
```diff
@@ -0,0 +1,94 @@
...
+impl AgentConfig {
+    /// Build the configuration from environment variables, applying defaults.
+    pub fn from_env() -> Result<Self> {
+        let gateway = std::env::var(ENV_GATEWAY).unwrap_or_else(|_| DEFAULT_GATEWAY.to_string());
+
+        let runner_dir = std::env::var_os(ENV_RUNNER_DIR).map(PathBuf::from);
+
+        let max_concurrent = match std::env::var(ENV_MAX_CONCURRENT) {
+            Ok(v) => v
+                .parse()
```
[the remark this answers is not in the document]
PeronGH: Fixed. `from_env` now rejects `ARCBOX_FLEET_MAX_CONCURRENT=0` explicitly with a clear error, so it can no longer parse to a 0 that makes the `len >= max_concurrent` check reject every dispatched job.

fleet/arcbox-fleet-agent/src/credentials.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// permissions to the owner.
+    pub fn store(&self, cred: &Credential) -> Result<()> {
+        if let Some(parent) = self.path.parent() {
+            std::fs::create_dir_all(parent)
+                .with_context(|| format!("creating {}", parent.display()))?;
+        }
+        let json = serde_json::to_vec_pretty(cred).context("serializing credential")?;
+        std::fs::write(&self.path, &json)
+            .with_context(|| format!("writing {}", self.path.display()))?;
+        restrict_permissions(&self.path)?;
```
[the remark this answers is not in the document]
PeronGH: Fixed. `store` now writes to a `*.json.tmp` and renames into place (atomic on the same filesystem), so a crash mid-write can no longer leave a truncated `credentials.json`. Permissions are restricted on the temp file before the rename, so the final path is never briefly world-readable.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /// Drop bookkeeping for a job, releasing its capacity slot.
+    fn release(&self, job_id: &str) {
+        self.inner.in_flight.remove(job_id);
+        self.inner.cancels.remove(job_id);
+    }
+
+    /// Emit a `RunnerFailed` and release the slot.
+    async fn fail(&self, job_id: &str, reason: &str) {
```
AprilNEA, line 251: Change requested: `RunnerFailed` currently only maps to the gateway/worker `finished` hint path, which releases capacity but does not terminally fail the runner job. That is fine for a runner that already started and will produce an authoritative GitHub completion webhook, but it is not fine for pre-start failures (`host is draining`, `host at capacity`, `failed to spawn runner`). In those cases no GitHub runner may ever pick up the job, so the workflow can sit in `Provisioning` until the completion deadline.

Can we split the semantics here? For pre-start/local startup failure, either requeue/try another machine or call a workflow path that marks the job failed while it is still in `Provisioning`. Keep the existing `finished` hint semantics only for post-start capacity release.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -0,0 +1,275 @@
...
+            return Admission::Draining;
+        }
+        if self.inner.in_flight.len() >= self.inner.max_concurrent {
+            return Admission::AtCapacity;
+        }
+        Admission::Accept
+    }
+
+    /// Cancel an in-flight job: aborting the task drops the child (kill-on-drop).
+    pub fn handle_cancel(&self, job_id: &str) {
```
AprilNEA, line 139: Change requested: canceling by aborting the task skips the normal `run_job` cleanup path, so the agent does not emit `RunnerFinished`/`RunnerFailed` and the server-side capacity slot can remain reserved until webhook/deadline. `kill_on_drop(true)` also only covers the immediate child (`run.sh` / `cmd /C run.cmd`), not necessarily the full runner/job process tree.

Please make cancellation an explicit path owned by `run_job`: store a cancellation token/channel per job, `select!` between `child.wait()` and cancellation, terminate the runner process tree, emit a terminal/canceled event, and then release local bookkeeping.

fleet/arcbox-fleet-agent/src/attach.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    let heartbeat = spawn_heartbeat(req_tx.clone(), config.max_concurrent);
+
+    let outcome = loop {
+        tokio::select! {
+            event = egress_rx.recv() => match event {
+                Some(msg) => {
+                    if let Err(err) = req_tx.send(msg).await {
```
AprilNEA, line 155: Change requested: this still treats enqueueing into the per-connection local `req_tx` as durable delivery. Once this send succeeds, the event is removed from the cross-reconnect queue, but it may still be lost if the connection drops before tonic writes it or before the gateway handles/relays it. That leaves terminal events/capacity release dependent on the webhook/deadline path.

To support the durability claim, we need an application-level ack/replay loop: keep lifecycle events in an unacked queue keyed by `(job_id, event_type)` until the gateway acks after the Restate handler succeeds, and replay unacked events after reconnect. The server handlers are already mostly idempotent enough for `started`/`finished` replay.

AprilNEA (changes requested):
Thanks for the iteration here — the agent shape is good, but I think we still need to close a few lifecycle/durability gaps before merging. The current implementation can still leave runner jobs/capacity stuck until the backstop deadline in common failure/reconnect paths.

AprilNEA:
Question (non-blocking): should the long-lived machine credential live in the OS credential store instead of `~/.arcbox/fleet/credentials.json`, at least on macOS? Since this agent is cross-platform, one option is the `keyring` crate (`keyring = "4.1.2"`, MIT OR Apache-2.0; docs: https://docs.rs/keyring/4.1.2), whose default backend covers Apple Keychain, Windows Credential Manager, and Linux Secret Service. That would avoid inventing per-platform file permission/atomic-replace behavior for the machine token. If we intentionally want a file for headless/server installs, maybe make that an explicit fallback or mode rather than the only storage path.

fleet/arcbox-fleet-agent/src/runner.rs
PeronGH: Fixed in 09067095.

Cancellation is now cooperative and owned by `run_job`: `cancels` stores a `oneshot::Sender<()>` instead of an `AbortHandle`, `handle_cancel` fires the signal, and `run_job` `select!`s between `child.wait()` and that signal. On cancel it SIGKILLs the **whole runner process group** — via `command-group` (POSIX process groups on Unix, a Job Object on Windows), so `run.sh`/`run.cmd` and every process it spawned go down, not just the immediate child — reaps it, then emits a terminal `RunnerFailed` and releases the slot through the normal path.

So both halves you raised are closed: the full process tree is torn down, and a terminal event is always emitted on cancel (the old task-abort path emitted none). Dropping `kill_on_drop` is safe because no path aborts the `run_job` task anymore, so it always runs to completion and reaps the group leader explicitly.

fleet/arcbox-fleet-agent/src/runner.rs
PeronGH: Being closed by the offer/reject redesign on the stacked PR #339 (RUN-23) rather than retrofitted here.

That redesign drops the terminal-event model entirely: the agent becomes the admission authority and answers each offer with `RunnerAccepted`/`RunnerRejected`. Every pre-start failure (draining, at-capacity, no matching capability, spawn/container-start failure) becomes a `reject`, and the platform `provision_runner` workflow re-offers to the next candidate; once candidates are exhausted the job fails fast — bounded by the 60s offer timeout × placement rounds, never the 6h completion deadline. I verified this end-to-end against the gateway and worker.

Since #310's `RunnerFailed`/`finished`-hint path is replaced wholesale in #339, splitting pre-start vs post-start semantics here would be churn on code that's about to be removed. Flagging so it's explicit — if you'd rather see the split land in #310 first, say the word and I'll do it here instead.

fleet/arcbox-fleet-agent/src/attach.rs
PeronGH: Also closed structurally in #339 (RUN-23) rather than here.

In that redesign the agent no longer emits terminal lifecycle events at all — it only answers offers with accept/reject tied to an `offer_token`. Durability comes from the gateway's at-least-once re-offer + idempotent duplicate handling, single-use GitHub JIT configs, and the authoritative `workflow_job=completed` webhook. I traced the lost-message paths against the platform repo:

- A dropped verdict leaves the per-offer awakeable unresolved → it times out (60s) and the workflow re-offers.
- A lost accept re-offers to another machine; the duplicate runner fails JIT registration (single-use), and the real run completes via the webhook.

So a lost event degrades to a bounded re-offer (≤ offer timeout), never the 6h backstop — which is exactly the guarantee the ack/replay loop was meant to provide. On #310's current model the gap you flagged is real, but it's the model #339 removes, so we're closing it there instead of building ack/replay into code that's being replaced.

fleet/arcbox-fleet-agent/src/credentials.rs
```diff
@@ -0,0 +1,107 @@
...
+                .with_context(|| format!("creating {}", parent.display()))?;
+        }
+        let json = serde_json::to_vec_pretty(cred).context("serializing credential")?;
+        // Write-then-rename so a crash mid-write can't leave a truncated
+        // credentials.json that fails to parse and strands the host unenrolled.
+        // Permissions are restricted on the temp file before the rename, so the
+        // final path is never briefly world-readable.
+        let tmp = self.path.with_extension("json.tmp");
+        std::fs::write(&tmp, &json).with_context(|| format!("writing {}", tmp.display()))?;
+        restrict_permissions(&tmp)?;
```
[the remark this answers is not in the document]
PeronGH: Confirmed and fixed in 14964b1b.

Verified the window: `std::fs::write` created the temp file at the umask default (0644) and only chmod'd it to 0600 afterward, and the parent dir is 0755 — so the raw machine token was briefly readable by any local process between the write and the chmod.

The temp file is now created 0600 in the `open(2)` call via `OpenOptionsExt::mode`, with an `fchmod` backstop on the open fd to cover a stale temp left by a crashed run (where `mode` is ignored on an existing file). Non-Unix keeps the `std::fs::write` fallback since ACL hardening is deferred. Thanks for the catch.

PeronGH:
Re the pullfrog review of 09067095 — confirming its read is correct: cooperative cancel owned by `run_job`, full process-group teardown, a terminal event always emitted, and no remaining task-abort path (so dropping `kill_on_drop` is safe — the task always runs to completion and reaps the group leader).

On the scoping note: the two items it correctly flags as still open — pre-start vs post-start failure semantics, and durable lifecycle-event delivery — are being closed by the offer/reject redesign in the stacked PR #339 (RUN-23), which replaces the terminal-event model both concerns are predicated on, rather than in #310. Details in the inline replies above.

lucas77778 (dismissed):

fleet/arcbox-fleet-agent/src/attach.rs
AprilNEA, line 155: This only proves the event was accepted into the local request channel, not that the gateway received or persisted it. If the h2/TCP stream dies after this send succeeds but before the gateway handles a `RunnerFinished`/`RunnerFailed`, the event is lost because it has already been removed from `egress_rx` and is not kept in `pending`. The current protocol needs an ACK/event-id or terminal-state replay mechanism before we can claim durable runner events across reconnects.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -0,0 +1,294 @@
...
+        tokio::spawn(async move { sup.run_job(order, cancel_rx).await });
+    }
+
+    /// Decide whether to start a runner for `job_id`. `Duplicate` takes
+    /// precedence over `Draining`/`AtCapacity`: a redelivery of an
+    /// already-running job is a no-op regardless of host state. The
+    /// `contains`/`insert` gap back in [`Self::handle_provision`] is safe
+    /// because the attach loop dispatches orders one at a time, with no
+    /// `.await` between this check and the reservation.
+    fn admit(&self, job_id: &str) -> Admission {
```
AprilNEA, line 119: This dedups only currently in-flight jobs. Once `release()` removes the job, a redelivered `ProvisionRunner` for the same `job_id` will be accepted and spawn another runner, which is possible if the terminal event was lost or not yet recorded by the gateway. Please retain a bounded completed/terminal outcome cache and replay the terminal outcome on duplicate redelivery instead of starting a new runner.

fleet/arcbox-fleet-agent/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+    command: Command,
+}
+
+#[derive(Debug, Subcommand)]
+enum Command {
+    /// Exchange an enrollment token for the machine credential and persist it.
+    Enroll {
+        /// Fleet enrollment token issued by the control plane (a multi-use
+        /// join token valid until it expires or is rotated).
+        #[arg(long)]
```
AprilNEA: This token is documented as a multi-use join token valid until expiry/rotation, so passing it only via argv is risky: it can leak through shell history, process listings, service manager logs, and support bundles. Please support a non-argv path such as stdin or `--token-file` and make that the recommended enrollment flow.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -0,0 +1,294 @@
...
+        self.inner
+            .draining
+            .store(true, std::sync::atomic::Ordering::Relaxed);
+        info!("draining: no new runners will be accepted");
+    }
+
+    /// Drive one runner process end to end, emitting accept/start/terminal
+    /// events. The runner is spawned as a process group so cancellation can take
+    /// down `run.sh` and every process it spawned, not just the immediate child.
+    async fn run_job(&self, order: ProvisionRunner, mut cancel_rx: oneshot::Receiver<()>) {
```
AprilNEA, line 196: The cancel path handles gateway `CancelRunner`, but there is no process-wide shutdown coordination. On SIGTERM/Ctrl-C/service stop, active runner process groups can be left behind because `main` just blocks on the infinite attach loop. Please add graceful shutdown that stops accepting work, signals/kills all active process groups, waits/reaps them with a timeout, and reports terminal/canceled state best-effort.

fleet/arcbox-fleet-agent/src/credentials.rs
```diff
@@ -0,0 +1,122 @@
...
+    // `mode` only takes effect when this call creates the file; a leftover temp
+    // from a crashed run keeps its old mode, so fchmod the open handle as a
+    // backstop (operating on the fd, never racing a path lookup).
+    file.set_permissions(std::fs::Permissions::from_mode(0o600))
+        .with_context(|| format!("chmod 0600 {}", path.display()))?;
+    file.write_all(bytes)
+        .with_context(|| format!("writing {}", path.display()))
+}
+
+/// Windows ACL hardening is deferred to a later slice.
```
AprilNEA: For a cross-platform agent storing a long-lived machine token, shipping Windows with plain `std::fs::write` and deferred ACL hardening is too risky. Please either restrict the DACL/use DPAPI or explicitly narrow the supported secure platforms until Windows credential storage is hardened.

AprilNEA (changes requested):
Thanks for pushing this forward — the standalone Fleet agent shape is good, and the later commits clearly address several real reconnect/cancel/security issues. I think this still needs another pass before merging because the PR currently promises durable/reconnect-safe runner reporting, but the protocol/client state machine cannot guarantee that yet.

Main blockers:

1. Runner lifecycle events are only queued into tonic, not acknowledged by the gateway, so terminal events can still be lost after local send succeeds but before the gateway records them.
2. Duplicate dispatch handling only covers in-flight jobs; a redelivery after local completion but before the terminal event is recorded can spawn the same job again.
3. Normal agent shutdown has no coordination to cancel/kill/reap active runner process groups.
4. The multi-use enrollment token is accepted only through argv, which leaks through shell history/process listings.
5. Windows credential storage is explicitly not hardened despite the agent being cross-platform.

Suggested path: add event ACK/terminal replay semantics first, retain completed-but-unacked job outcomes for duplicate redelivery, add graceful shutdown for active process groups, and move enrollment token input to stdin/file/env-secret before landing.

PeronGH:
@AprilNEA thanks for the thorough pass. Pushed three commits that close the blockers which belong to this PR's base-agent scope, and I want to be explicit about the two that I'm intentionally **not** fixing here. Mapping to your numbered list:

**3. Graceful shutdown of active runners — fixed in aede79b.**
The agent blocked on the infinite attach loop with no signal handling. There's now a termination-signal listener (Ctrl-C on any platform, SIGTERM on Unix) that cancels a token threaded through the reconnect loop. On shutdown the loop exits and `RunnerSupervisor::shutdown` drains (stops accepting offers), fires the cancel signal for every in-flight job — each `run_job` SIGKILLs and reaps its whole process group — then waits, bounded by a 15s grace, for the slots to release. So no runner process group is left orphaned on service stop.
Honest caveat: terminal events on shutdown are **best-effort** (the stream may already be torn down). The guarantee is "no orphaned process groups"; the gateway reclaims capacity when it observes the dropped attach stream.

**4. Enrollment token only through argv — fixed in 0f018db.**
`enroll` now reads the join token from `--token-file` (recommended) or stdin by default. `--token` is kept for convenience but is documented as insecure and `conflicts_with` `--token-file`. The token is trimmed and rejected if empty.

**5. Windows credential storage — narrowed to Unix in 378710b.**
Rather than ship untested `unsafe` Win32 DACL code from a macOS/Linux dev box, the non-Unix path now **refuses** to persist the token instead of writing it at the default ACL. Storage is Unix-only (`0600`) until proper DACL hardening lands on a platform we can actually test. That removes the silent-insecure path; the DACL implementation is tracked as follow-up.

**1. Lifecycle-event ACK/replay** and **2. duplicate dispatch after local completion** — these are closed in #339 (RUN-23), not here. They are both predicated on the terminal-event model, which #339 deletes wholesale: the agent becomes the admission authority, answers offers with accept/reject keyed by `offer_token`, and resends each verdict until the gateway sends an `OfferVerdictAck` — the application-level ack/replay you asked for, already implemented there. Retrofitting ack/replay into `RunnerFinished`/`RunnerFailed` in #310 would be throwaway code on the exact model the next PR in the stack removes. So these stay tracked against #339.

AprilNEA:
I think the long-lived machine credential should default to the OS credential store via a cross-platform crate rather than a file in `~/.arcbox/fleet/credentials.json`.

For a cross-platform agent, using a common abstraction is the cleaner default:
- macOS → Keychain
- Windows → Credential Manager
- Linux → Secret Service

That avoids us owning per-platform file permission / atomic replace / ACL hardening details for the sensitive path. A file-based store still makes sense, but more as an explicit fallback for headless/server environments rather than the primary storage mode.

So my preference would be:
1. default to a cross-platform keychain/keyring crate
2. keep file storage as an opt-in or clearly defined fallback mode
3. narrow the supported fallback story explicitly where the platform cannot provide secure storage guarantees yet

AprilNEA (approved):

PR #316, opened 2026-06-17
feat(splicetcp): parameterize FrameClassifier packet-pool capacity

AprilNEA:
## What

`FrameClassifier` hardcoded a 4096-buffer `PacketPool`. Each `PacketBuffer` is a fixed `[u8; MAX_PACKET_SIZE]` (64 KiB), so the pool is **~256 MiB resident** — sized for the VM datapath (one pool per VM, compressible idle pages, burst headroom at 10 Gbps).

This change makes the pool capacity a constructor parameter:

- **`FrameClassifier::with_pool_capacity(gateway_ip, mtu, pool_capacity)`** — explicit buffer count.
- **`FrameClassifier::DEFAULT_POOL_CAPACITY`** (`4096`) — now a `pub const`.
- **`new()` delegates** to `with_pool_capacity` with the default, so **every existing call site is unchanged** (non-breaking).

Also corrects a stale comment: `PacketPool::new` doesn't "only fail on zero capacity" — it clamps capacity to `>= 1` (`capacity.max(1)`) and is infallible in practice.

## Why

The 256 MiB pool blows the **~50 MiB memory cap of an iOS Network Extension**, blocking a memory-constrained NE consumer of the classifier/bridge datapath. Such a host can now construct with a small count (e.g. `256` ⇒ ~16 MiB).

The pool only bounds **burst headroom**: once exhausted the classifier falls back to heap allocation, so a small pool trades burst throughput for a bounded footprint — **never correctness**.

## Test

`small_pool_capacity_classifies_and_falls_back_to_heap`: a `cap = 4` classifier ingests `cap + 2` frames with none drained, and all land in the rx queue — proving the heap fallback past capacity does not drop frames.

## Verification

- `cargo test -p splicetcp` — 34 passed, 0 failed
- `cargo fmt --all --check` — clean
- `cargo clippy -p splicetcp --all-targets -- -D warnings` — clean
- `cargo check --workspace` — clean (all `FrameClassifier::new` call sites unaffected)

### DIFF
--- a/common/splicetcp/src/classifier.rs
+++ b/common/splicetcp/src/classifier.rs
@@ -107,15 +107,25 @@ pub struct FrameClassifier {
 }
 
 impl FrameClassifier {
-    /// Pool capacity: 4096 buffers. Each buffer is `MAX_PACKET_SIZE` (64 KiB)
-    /// in the pool, so total resident = ~256 MiB. This is acceptable because:
+    /// Default packet-pool capacity: 4096 buffers. Each buffer is
+    /// `MAX_PACKET_SIZE` (64 KiB), so resident pool memory is ~256 MiB. This is
+    /// acceptable for the VM datapath because:
     /// - One pool per VM (not per connection)
     /// - Idle pages are compressed by macOS memory compressor
     /// - Smaller pools increase heap fallback frequency, defeating the purpose
     /// - At ~300K frames/sec (10 Gbps / 4000 MTU), 4096 gives ~13 ms burst headroom
-    const POOL_CAPACITY: usize = 4096;
+    ///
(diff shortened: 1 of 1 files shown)


AprilNEA:
Addressed the P2 (naming/value divergence): the value stays `4096` intentionally — it preserves `FrameClassifier`'s historical pool size, and bumping to `arcbox_datapath::DEFAULT_POOL_CAPACITY` (8192) would silently double the VM datapath's classifier-pool memory, defeating this PR's non-breaking goal. Since the value must differ, I documented the divergence on the constant explicitly (it's an associated const — `FrameClassifier::DEFAULT_POOL_CAPACITY` — so the call site is already namespaced) rather than renaming. (`ad6e473`)

PR #319, opened 2026-06-17
ci(release): publish splicetcp datapath cluster to crates.io on tag

AprilNEA:
## What

Adds a `publish-crates` job to the release-build workflow that publishes the workspace's reusable datapath core to crates.io on every release tag.

The repo uses **release-please** (which only opens release PRs + tags) — there was no `cargo publish` step anywhere, so tagged releases never reached crates.io. This wires that in.

## Which crates

The five-crate splicetcp cluster, published as one coordinated `cargo publish`:

```
arcbox-packet    (leaf)
arcbox-datapath  (leaf)
arcbox-fakeip    (leaf)
arcbox-proxy  -> arcbox-packet, arcbox-fakeip
splicetcp     -> arcbox-{packet,datapath,fakeip,proxy}
```

## Behavior

- Runs on each `v*` tag, **in parallel** with the binary builds — the library crates are their own deliverable and must not be gated on Apple code-signing.
- `if: github.event_name == 'push'` — skipped on `workflow_dispatch` so manual binary rebuilds of an old tag don't republish.
- **One coordinated publish** via `cargo publish -p … (×5)` (cargo ≥ 1.90): cargo packages all five, verifies each against the others through a local temp registry (so the as-yet-unpublished inter-deps resolve), then uploads in dependency order, waiting for the index between uploads. No hand-rolled ordering, and no idempotency guard — cargo verifies every crate *before* uploading any, so there's no partial-failure window to guard against.
- `--workspace` is intentionally **not** used: it would also select the VM crates, several of which depend on `publish = false` members (e.g. `arcbox-dns`) and therefore can't be published.
- `--locked` — relies on the lockfile release-please refreshes on the release PR.

## Required before first release

Add a `CARGO_REGISTRY_TOKEN` repo secret, and confirm the `splicetcp` / `arcbox-*` crate names are free or owned on crates.io (first publish creates them).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,55 @@ jobs:
           echo "Building version: $VERSION"
 
   # ==========================================================================
+  # Publish the splicetcp datapath cluster to crates.io
+  # ==========================================================================
+  # These five crates are the only `publish = true` members of the workspace
+  # (everything else is `publish = false`): the reusable, open-source datapath
+  # core that downstream projects consume as published deps instead of a path
+  # into a local checkout. Independent of the binary builds — the library
+  # crates are their own deliverable and must not be gated on Apple signing.
+  # Skipped on workflow_dispatch (manual binary rebuilds must not republish).
+  publish-crates:
+    name: Publish crates to crates.io
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


AprilNEA:
Superseded by #321: 方向从'只发 splicetcp cluster(-p ×5)'改为'发布所有库 crate(cargo publish --workspace)'。#321 含完整 publish-crates job + 把 9 个库 crate 翻成可发布 + 闭包修复。

PR #320, opened 2026-06-18
chore: restrict crates.io publishing to the splicetcp cluster

AprilNEA:
## What

Marks every workspace member **except the splicetcp datapath cluster** as `publish = false`, so the crates.io publish set is closed under dependencies.

Publishable after this change — exactly the 5 crates `release.yml` publishes (#319):

```
splicetcp, arcbox-proxy, arcbox-fakeip, arcbox-packet, arcbox-datapath
```

## Why

The publish flags were inconsistent: 28 members were publishable, but two of them — `arcbox-net` and `arcbox-vmm` — depend on `arcbox-dns`, which is `publish = false`. crates.io forbids a path/version dep on an unpublished crate, so those two could **never actually be published**, and a bare `cargo publish --workspace` aborts at `arcbox-net → arcbox-dns`. The rest of the VM crate family (`arcbox-virtio*`, `arcbox-vm`, `arcbox-hv`, `arcbox-vz`, …) was publishable-by-default but nothing publishes it.

This pins intent: only the splicetcp cluster (consumed downstream by aroxy) goes to crates.io.

## Effect

- Publishable members: **5** (was 28); each one's dependency closure now contains only publishable crates (verified via `cargo metadata`).
- `arcbox-net` / `arcbox-vmm` → `arcbox-dns` inconsistency resolved.
- **No build/runtime impact** — `publish = false` only gates crates.io uploads; path deps within the workspace and aroxy are unaffected.

Independent of #319 (which already publishes only the 5 via an explicit `-p` allowlist); this makes the workspace config match that intent.

### DIFF
--- a/app/arcbox-migration/Cargo.toml
+++ b/app/arcbox-migration/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "arcbox-migration"
+publish = false
 version.workspace = true
 edition.workspace = true
 rust-version.workspace = true
--- a/common/arcbox-conntrack/Cargo.toml
+++ b/common/arcbox-conntrack/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "arcbox-conntrack"
+publish = false
 description = "Stateful NAT connection tracking and zero-copy packet translation for ArcBox"
 version.workspace = true
 edition.workspace = true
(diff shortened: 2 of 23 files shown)


AprilNEA:
改变方向:不再把其余 crate 限制成 publish=false,改为'能发布的都发布'。关闭本 PR(未合并,master 未受影响),另行处理。

PR #330, opened 2026-06-25
fix(daemon): reconcile vmnet routes after restart

AprilNEA:
## Summary
- Reconcile the container subnet route through the vmnet bridge during daemon cold-start recovery
- Detect arcbox-helper/daemon version drift during self-setup

### DIFF
--- a/app/arcbox-cli/src/commands/doctor.rs
+++ b/app/arcbox-cli/src/commands/doctor.rs
@@ -45,7 +45,7 @@ pub async fn execute() -> Result<()> {
     {
         check!("Bridge NIC", check_bridge_nic());
         check!("Container route", check_container_route());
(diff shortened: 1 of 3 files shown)


AprilNEA:
Addressed the AI review test coverage note by adding focused unit coverage for cold-start route plan selection under vmnet/non-vmnet cfgs. Kept the duplicated version-drift comparison as-is because daemon self-setup and doctor have different output/reporting responsibilities and the current duplication is small.\n\nRe-verified:\n- `cargo fmt --check`\n- `cargo check -p arcbox-daemon`\n- `cargo clippy -p arcbox-daemon -p arcbox-cli -- -D warnings`\n- `cargo test -p arcbox-daemon`

PR #334, opened 2026-06-25
fix(storage): drop upfront docker.img preallocation, keep image sparse

AprilNEA:
## Problem

On first VM creation the data image (`docker.img`) physically pre-allocated up to **64 GiB** on disk via the macOS `F_PREALLOCATE` path. A fresh install therefore reported tens of GiB of disk usage with **zero containers** — wasteful, and a regression against OrbStack's thin data image. Reported in arcboxlabs/arcbox-desktop#244 (ABXD-95).

## Fix

Remove the `F_PREALLOCATE` block entirely and rely on `set_len`, which extends only the logical size and leaves the file sparse. APFS/Btrfs allocate on write lazily, so the working set still gets CoW behaviour without the upfront physical cost. An existing image is still never shrunk.

After the fix, an empty install reports negligible on-disk usage; the image only consumes space for blocks the guest actually writes.

## Tests

- `ensure_sparse_block_image_is_thin_provisioned` — asserts the created image reports its full virtual size yet allocates < 1 MiB of physical blocks (regression guard against any reintroduced preallocation).
- `ensure_sparse_block_image_never_shrinks` — a subsequent smaller-size call must not truncate an existing image; a larger size grows it.

`cargo build` / `cargo test` / `cargo clippy --all-targets` / `cargo fmt` all clean.

### DIFF
--- a/app/arcbox-core/src/vm_lifecycle/mod.rs
+++ b/app/arcbox-core/src/vm_lifecycle/mod.rs
@@ -351,6 +351,19 @@ pub struct VmLifecycleManager {
 }
 
 impl VmLifecycleManager {
+    /// Ensures a sparse, thin-provisioned block image of `size_bytes` virtual
+    /// size exists at `path`, creating parent directories as needed.
+    ///
+    /// `set_len` extends only the logical size (EOF); it reserves no physical
+    /// blocks, so the host file stays sparse and consumes disk only for blocks
+    /// the guest actually writes — matching OrbStack's thin data image.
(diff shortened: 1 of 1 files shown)


AprilNEA:
Follow-up stacked on this branch: #336 — adds host-side DISCARD hole-punching so guest `fstrim` actually reclaims disk (including the previously preallocated space on existing installs), plus wires up `arcbox disk compact`. Merge this PR first.

PR #339, opened 2026-06-25
feat(fleet): offer/reject protocol, agent as admission authority (RUN-23, client side)

PeronGH:
Client half of the RUN-23 fleet protocol redesign. Pairs with the platform PR (`arcboxlabs/platform#100`); both repos move to `protocol_version = 1` together.

Stacked on #328 (`feat/fleet-docker`) — retarget to `master` once #328 merges.

## What changed

The agent is now the **admission authority** under the new offer/reject protocol; the server holds no capacity model.

- **proto** re-vendored: `Capability(os, arch, backed_by)` + `HostTelemetry`; `ProvisionRunner` carries an `offer_token`; `AttachRequest` is `Heartbeat`/`RunnerAccepted`/`RunnerRejected`; `EnrollRequest` carries `protocol_version`.
- **host**: advertise capabilities (`host_runner` / `docker` per `(os, arch)`; darwin is `host_runner` until the VM backend lands) and report live telemetry (load-avg, cpu, total/available memory).
- **runner**: per-offer admission — draining / capability routing / live load + memory headroom; a redelivered offer for a running job is an idempotent accept. **Starts the runner, then accepts**; any failure up to start → reject so the platform re-offers. Backend chosen by capability lookup (the `if os == "linux"` inference is gone). No terminal events — the GitHub webhook is authoritative.
- **docker**: split run into `start` + `RunningContainer::wait`, so accept is sent only once the container is actually running; dropped the image override.
- **config**: dropped `ARCBOX_FLEET_MAX_CONCURRENT`; added `ARCBOX_FLEET_LOAD_CEILING` (0.9) + `ARCBOX_FLEET_MEM_FLOOR_MIB` (2048) + `PROTOCOL_VERSION = 1`; attach sends the version handshake.

## Verification

Fleet-scoped gate passes: `cargo clippy -p arcbox-fleet-agent -p arcbox-fleet-proto --all-targets -- -D warnings`, `cargo fmt --check`, `cargo test -p arcbox-fleet-agent` (10/10). (Workspace clippy can't run on Linux — macOS-only crates need `cc1obj`.)

**Fleet E2E (no mocks) — ✅ PASSED on the dev box (2026-06-26).** This agent, built and run natively on the box against the new gateway (platform `protocol_version = 1`):

enroll (version handshake) → attach (advertises `linux/amd64` `docker`, reports telemetry) → on a real `runs-on: [self-hosted, linux, X64]` job: gateway offer → `docker.start` → `RunnerAccepted` resolved the workflow's per-offer awakeable → the runner executed the job inside the actions-runner container → GitHub `workflow_job=completed` drove the terminal state. Job completed **success** end-to-end; output confirmed execution on the fleet. (The reject/re-offer path remains covered by `runner.rs` unit tests, not forced live.)

## Still pending

- [ ] Retarget to `master` after #328 merges.
- [x] The vendored `fleet.proto` is a **manual copy** (pre-publish). After the platform proto publishes to the BSR, run `make fleet-proto-sync` to replace it with the canonical module.

### DIFF
--- a/fleet/arcbox-fleet-agent/src/config.rs
+++ b/fleet/arcbox-fleet-agent/src/config.rs
@@ -14,13 +14,20 @@ pub const DEFAULT_GATEWAY: &str = "https://fleet.arcbox.dev";
 
 const ENV_GATEWAY: &str = "ARCBOX_FLEET_GATEWAY";
 const ENV_RUNNER_DIR: &str = "ARCBOX_FLEET_RUNNER_DIR";
-const ENV_MAX_CONCURRENT: &str = "ARCBOX_FLEET_MAX_CONCURRENT";
+const ENV_LOAD_CEILING: &str = "ARCBOX_FLEET_LOAD_CEILING";
+const ENV_MEM_FLOOR_MIB: &str = "ARCBOX_FLEET_MEM_FLOOR_MIB";
 const ENV_DATA_DIR: &str = "ARCBOX_FLEET_DATA_DIR";
 const ENV_DOCKER: &str = "ARCBOX_FLEET_DOCKER";
 const ENV_RUNNER_IMAGE: &str = "ARCBOX_FLEET_RUNNER_IMAGE";
 
-const DEFAULT_MAX_CONCURRENT: usize = 2;
+/// Reject an offer when 1-minute load average per core exceeds this.
+const DEFAULT_LOAD_CEILING: f64 = 0.9;
+/// Reject an offer when available memory is below this many MiB.
+const DEFAULT_MEM_FLOOR_MIB: u64 = 2048;
 const DEFAULT_RUNNER_IMAGE: &str = "ghcr.io/actions/actions-runner:latest";
+
+/// Wire-protocol version this agent speaks; the gateway rejects a mismatch.
+pub const PROTOCOL_VERSION: u32 = 1;
 
 /// Whether Docker-based Linux job execution is enabled.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -37,8 +44,7 @@ pub enum DockerMode {
 #[derive(Debug, Clone)]
 pub struct DockerConfig {
     pub mode: DockerMode,
-    /// Container image used for Linux runner jobs when the platform sends no
-    /// image (empty `ProvisionRunner.image`).
+    /// Container image used for Linux runner jobs.
     pub runner_image: String,
 }
 
@@ -50,8 +56,10 @@ pub struct AgentConfig {
     /// Directory holding the pre-installed GitHub Actions runner (`run.sh`/`run.cmd`).
     /// `None` until set; required only by the `run` command.
     pub runner_dir: Option<PathBuf>,
-    /// Maximum number of concurrently executing runner jobs.
-    pub max_concurrent: usize,
+    /// Reject an offer when 1-minute load average per core exceeds this.
+    pub load_ceiling: f64,
+    /// Reject an offer when available memory (MiB) is below this.
+    pub mem_floor_mib: u64,
     /// Agent data directory (credentials, logs). Defaults to `~/.arcbox/fleet`.
     pub data_dir: PathBuf,
     /// Docker runtime configuration for Linux jobs.
@@ -65,17 +73,17 @@ impl AgentConfig {
 
         let runner_dir = std::env::var_os(ENV_RUNNER_DIR).map(PathBuf::from);
 
-        let max_concurrent = match std::env::var(ENV_MAX_CONCURRENT) {
-            Ok(v) => {
-                let n: usize = v
-                    .parse()
-                    .with_context(|| format!("{ENV_MAX_CONCURRENT} must be a positive integer"))?;
-                if n == 0 {
-                    anyhow::bail!("{ENV_MAX_CONCURRENT} must be a positive integer, got 0");
-                }
-                n
-            }
-            Err(_) => DEFAULT_MAX_CONCURRENT,
+        let load_ceiling = match std::env::var(ENV_LOAD_CEILING) {
+            Ok(v) => v
+                .parse()
+                .with_context(|| format!("{ENV_LOAD_CEILING} must be a number"))?,
+            Err(_) => DEFAULT_LOAD_CEILING,
+        };
+        let mem_floor_mib = match std::env::var(ENV_MEM_FLOOR_MIB) {
+            Ok(v) => v
+                .parse()
+                .with_context(|| format!("{ENV_MEM_FLOOR_MIB} must be a non-negative integer"))?,
+            Err(_) => DEFAULT_MEM_FLOOR_MIB,
         };
 
         let data_dir = match std::env::var_os(ENV_DATA_DIR) {
@@ -97,7 +105,8 @@ impl AgentConfig {
         Ok(Self {
             gateway,
             runner_dir,
-            max_concurrent,
+            load_ceiling,
+            mem_floor_mib,
             data_dir,
             docker: DockerConfig {
                 mode: docker_mode,
--- a/fleet/arcbox-fleet-agent/src/config.rs
+++ b/fleet/arcbox-fleet-agent/src/config.rs
@@ -74,9 +74,15 @@ impl AgentConfig {
         let runner_dir = std::env::var_os(ENV_RUNNER_DIR).map(PathBuf::from);
 
         let load_ceiling = match std::env::var(ENV_LOAD_CEILING) {
-            Ok(v) => v
-                .parse()
-                .with_context(|| format!("{ENV_LOAD_CEILING} must be a number"))?,
+            Ok(v) => {
+                let n: f64 = v
+                    .parse()
+                    .with_context(|| format!("{ENV_LOAD_CEILING} must be a positive number"))?;
+                if n <= 0.0 {
+                    anyhow::bail!("{ENV_LOAD_CEILING} must be a positive number, got {n}");
+                }
+                n
+            }
             Err(_) => DEFAULT_LOAD_CEILING,
         };
         let mem_floor_mib = match std::env::var(ENV_MEM_FLOOR_MIB) {
--- a/fleet/arcbox-fleet-agent/src/config.rs
+++ b/fleet/arcbox-fleet-agent/src/config.rs
@@ -19,6 +19,7 @@ const ENV_MEM_FLOOR_MIB: &str = "ARCBOX_FLEET_MEM_FLOOR_MIB";
 const ENV_DATA_DIR: &str = "ARCBOX_FLEET_DATA_DIR";
 const ENV_DOCKER: &str = "ARCBOX_FLEET_DOCKER";
 const ENV_RUNNER_IMAGE: &str = "ARCBOX_FLEET_RUNNER_IMAGE";
+const ENV_CREDENTIAL_STORE: &str = "ARCBOX_FLEET_CREDENTIAL_STORE";
 
 /// Reject an offer when 1-minute load average per core exceeds this.
 const DEFAULT_LOAD_CEILING: f64 = 0.9;
@@ -38,6 +39,19 @@ pub enum DockerMode {
     Enabled,
     /// Never use Docker, even if available.
     Disabled,
+}
+
+/// Where the long-lived machine credential is persisted.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CredentialMode {
+    /// OS keychain on macOS (login Keychain) and Windows (Credential Manager);
+    /// the owner-only data-dir file on Linux. On macOS the agent must run in the
+    /// user's session, where the login Keychain is unlocked.
+    Auto,
+    /// Force the OS keychain. Supported on macOS and Windows; errors on Linux.
+    Keyring,
+    /// Force the data-dir file (`0600` on Unix).
+    File,
 }
 
 /// Docker-specific configuration for running Linux jobs in containers.
@@ -64,6 +78,8 @@ pub struct AgentConfig {
     pub data_dir: PathBuf,
     /// Docker runtime configuration for Linux jobs.
     pub docker: DockerConfig,
+    /// Where the machine credential is persisted (OS keychain vs file).
+    pub credential_store: CredentialMode,
 }
 
 impl AgentConfig {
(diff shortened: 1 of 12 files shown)


AprilNEA:
resolve comment plz

fleet/arcbox-fleet-agent/src/host.rs
```diff
@@ -60,75 +60,65 @@ pub fn host_info_json() -> String {
...
+pub fn telemetry() -> HostTelemetry {
+    let mut sys = sysinfo::System::new();
+    sys.refresh_memory();
+    HostTelemetry {
+        load_avg_1m: sysinfo::System::load_average().one,
+        cpu_count: cpu_cores(),
+        mem_total_mib: sys.total_memory() / 1024 / 1024,
+        mem_available_mib: sys.available_memory() / 1024 / 1024,
+    }
 }
```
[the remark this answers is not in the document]
PeronGH: Declining this one. The allocation is negligible — `System::new()` is empty and `refresh_memory()` is just a couple of syscalls. A shared long-lived `System` would need synchronization since `telemetry()` is called concurrently from both the offer path and the heartbeat tick, which adds more complexity than the micro-optimization is worth. Keeping the per-call construction.

PeronGH:
@AprilNEA following up on the suggestion to keep the long-lived machine credential in the OS credential store rather than a flat file.

We added a keyring-backed `CredentialStore` on **macOS and Windows**. The backend is selected by `ARCBOX_FLEET_CREDENTIAL_STORE` (`auto` | `keyring` | `file`); `auto` = **OS keychain on macOS (login Keychain) and Windows (Credential Manager), owner-only `0600` file on Linux**. The `keyring` dependency is scoped to `cfg(any(target_os = "macos", target_os = "windows"))`, so Linux pulls none of it.

**Why it's a real improvement, not just at-rest encryption.** The keychain item is access-controlled to the agent binary by the OS — verified in `apple-native-keyring-store`: it stores via `SecKeychainAddGenericPassword` on the login keychain, whose default ACL trusts only the creating app. So a *different* same-user process — a malicious CI job, an unsigned binary — is prompted and, with nothing to click Allow, denied. The `0600` file has no such gate: any same-UID process reads it. That same-user boundary is what matters on a runner host. (On Windows it also turns the platform from *unsupported* into secure — the file backend was refused there because a default-ACL file is world-readable.)

**macOS caveat — it requires running in the user's session.** The login Keychain is unlocked at user login, so the agent must run as a post-login agent, not a session-less daemon. That's the intended deployment. The System keychain would remove the session requirement but needs root, which we don't run as — so the login Keychain (post-login agent) is the non-root path.

**Linux stays on the file.** Neither Linux backend fits a headless, reboot-persistent, unattended agent: kernel keyutils is wiped on reboot, and Secret Service needs a D-Bus session + a running gnome-keyring/KWallet + an unlock (auto-unlocking it unattended means storing the unlock secret in plaintext — the same problem). Encryption at rest on Linux is the deployment's job (LUKS).

Caveat for reviewers: the keychain backends are `cfg`-gated to macOS/Windows, so they aren't compiled by the Linux clippy/test path and need a macOS/Windows build to verify. The file backend, config, and wiring are covered by the existing Linux checks.

fleet/arcbox-fleet-agent/src/config.rs
```diff
@@ -65,17 +89,23 @@ impl AgentConfig {
...
+        let load_ceiling = match std::env::var(ENV_LOAD_CEILING) {
             Ok(v) => {
-                let n: usize = v
+                let n: f64 = v
                     .parse()
-                    .with_context(|| format!("{ENV_MAX_CONCURRENT} must be a positive integer"))?;
-                if n == 0 {
-                    anyhow::bail!("{ENV_MAX_CONCURRENT} must be a positive integer, got 0");
+                    .with_context(|| format!("{ENV_LOAD_CEILING} must be a positive number"))?;
+                if n <= 0.0 {
```
AprilNEA: This rejects zero and negative values, but still accepts `NaN`: `NaN <= 0.0` is false, and later `load_per_core > load_ceiling` is also false, disabling the load gate. Please require `n.is_finite() && n > 0.0` so `NaN`/`inf` are rejected at startup.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -340,63 +342,115 @@ impl RunnerSupervisor {
...
-        let image = docker.resolve_image(&order.image);
+        // Invariant: a Docker-backed capability is only advertised when Docker
+        // is present, so admit() routes here only with `docker` set.
+        let docker = self
+            .inner
+            .docker
+            .as_ref()
+            .expect("docker backend implies a docker runtime");
+
+        let running = match docker
```
AprilNEA: Cancellation is not observed while Docker startup is in progress. `handle_provision` installs the cancel sender before spawning the job task, but this path awaits the full `docker.start(...)` (pull/create/start) before entering the cancel `select!`. A slow image pull or container create can receive `CancelRunner`, then still start and later send `RunnerAccepted`. Startup should be cancellation-aware: race cancel against pull/create/start, clean up any partially-created container, and re-check cancellation before accepting the offer.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -1,171 +1,185 @@
...
-    /// orders one at a time, with no `.await` between this check and the
-    /// reservation.
-    fn admit(&self, job_id: &str, pool: &Pool) -> Admission {
+    /// Decide whether to run `job_id`. `Duplicate` takes precedence so a
+    /// redelivery is never a fresh runner. Then drain, then capability routing,
+    /// then live headroom (load per core and free memory). The check/insert gap
+    /// in [`Self::handle_provision`] is safe: the attach loop dispatches offers
+    /// one at a time with no `.await` between this check and the reservation.
+    fn admit(&self, job_id: &str, os: &str, arch: &str, telemetry: &HostTelemetry) -> Admission {
         if self.inner.in_flight.contains_key(job_id) {
```
AprilNEA: Duplicate detection only considers currently-running jobs. After a runner exits, `ReleaseGuard` removes `in_flight`, but an accepted verdict may still be unacked in `outstanding`. If the gateway redelivers that same offer before the ack arrives, this code will admit and start a second runner instead of replaying the existing verdict. `handle_provision` should first check `outstanding` by `offer_token` and requeue the exact stored verdict without re-admitting or starting work. If the gateway can redeliver the same job under a new token, keep a short-lived accepted-job record as well.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -340,63 +342,115 @@ impl RunnerSupervisor {
...
+        self.inner.outstanding.remove(offer_token);
+    }
+
+    /// Resend every still-unacked verdict onto the egress queue, bounded by
+    /// [`VERDICT_MAX_RESENDS`]. Non-blocking (`try_send`): a full buffer or a
+    /// momentarily detached stream just retries on the next tick. Driven by the
+    /// attach loop's resend ticker and, implicitly, every reconnect.
+    pub fn resend_outstanding(&self) {
+        self.inner.outstanding.retain(|token, entry| {
+            if entry.attempts >= VERDICT_MAX_RESENDS {
```
AprilNEA: Dropping every unacked verdict after the local resend cap is unsafe for `RunnerAccepted`. Once the agent has accepted, the platform needs that verdict to transition the job to running; the agent no longer sends terminal events, so silently expiring an accepted verdict can strand the protocol state. At minimum, do not locally drop accepted verdicts without an explicit gateway ack/expiry signal. If a cap is still desired, it should apply only where loss is safe, or only after the gateway tells the agent the offer is obsolete.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -340,63 +342,115 @@ impl RunnerSupervisor {
...
+    /// Record a verdict as outstanding (resent until acked) and send it once now.
+    async fn send_verdict(&self, token: &str, msg: attach_request::Msg) {
+        self.inner.outstanding.insert(
+            token.to_owned(),
+            Outstanding {
+                msg: msg.clone(),
+                attempts: 0,
+            },
+        );
+        self.send(msg).await;
```
AprilNEA: This await makes verdict transport backpressure part of runner supervision. Both host and Docker call `accept(...).await` before entering their wait/cancel loops, so if the egress queue is full or the attach loop is between streams/backing off, the runner task can block here and stop observing cancellation or reaping the runner/container. Since `outstanding` is already the reliability source of truth, verdict emission should record the outstanding verdict and then `try_send`; if the queue is full/closed, rely on the resend loop / next attach to deliver it.

AprilNEA (changes requested):
Thanks for the RUN-23 client-side redesign. The high-level shape is promising: the agent-side admission authority, accept-after-start semantics, reconnect-surviving egress queue, and credential hardening are all moving in the right direction.

I would still request changes before merging. There are several lifecycle/reliability edges where the implementation records local state as settled before the resource or protocol state is actually settled. Pullfrog already called out the Docker shutdown/orphan issue, so I did not duplicate that inline. The remaining blockers are around cancellation during startup, verdict delivery backpressure, duplicate offer handling after runner exit, and unacked accepted verdict expiry.

The structural fix I would look for is to make job/resource lifecycle and verdict lifecycle explicit and owned by one supervisor path: cancellation must be observed during startup, cleanup must be awaited before release, verdict enqueue must not block runner supervision, and accepted verdicts should remain outstanding until the gateway explicitly acknowledges or expires them.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -340,63 +342,115 @@ impl RunnerSupervisor {
...
-                Err(e) => self.fail(job_id, &format!("docker runner: {e}")).await,
+                Err(e) => warn!(job_id, error = %e, "waiting on container failed"),
             },
-            // CancelRunner: dropping `run` drops the container guard, which kills
-            // and removes the container.
+            // CancelRunner: dropping the wait future drops the container guard,
+            // which kills and removes the container.
             _ = &mut cancel_rx => {
-                self.fail(job_id, "canceled by gateway").await;
+                info!(job_id, "runner canceled");
```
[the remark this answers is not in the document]
PeronGH: Fixed in acc509de. Docker teardown is now an explicit awaited path: `RunningContainer::cancel()` kills and force-removes the container inline in the cancel branch, so `ReleaseGuard` clears `in_flight` only after the container is actually gone and `shutdown`'s grace loop bounds real teardown, matching the host path. `ContainerGuard` is demoted to a panic-only safety net and is no longer the path any guarantee rides on.

fleet/arcbox-fleet-agent/src/runner.rs
```diff
@@ -340,63 +342,115 @@ impl RunnerSupervisor {
...
+                warn!(offer_token = %token, "giving up on unacknowledged verdict");
+                return false;
+            }
+            entry.attempts += 1;
+            // Ignore send errors: outstanding stays, the next tick retries.
+            let _ = self.inner.events.try_send(AttachRequest {
+                msg: Some(entry.msg.clone()),
+            });
+            true
+        });
```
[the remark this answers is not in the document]
PeronGH: Fixed in 0b0fff78 by removing the attempt counter entirely — see the runner.rs:440 thread: verdicts are no longer expired locally, so there is nothing to miscount. `resend_outstanding` is now a plain re-emit of every unacked verdict via `try_send`, and a full buffer genuinely just retries next tick.

fleet/arcbox-fleet-agent/src/credentials.rs
```diff
@@ -86,31 +141,67 @@ fn write_private(path: &Path, bytes: &[u8]) -> Result<()> {
...
+/// Stores the credential as a single JSON blob in the OS keychain.
+#[cfg(any(target_os = "macos", target_os = "windows"))]
+pub struct KeyringStore;
+
+#[cfg(any(target_os = "macos", target_os = "windows"))]
+impl KeyringStore {
+    /// Keychain service and account under which the credential blob is stored.
+    const SERVICE: &'static str = "dev.arcbox.fleet-agent";
+    const ACCOUNT: &'static str = "machine-credential";
+
```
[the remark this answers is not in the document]
PeronGH: Fixed in 8dab0a4b. The keychain account is now the gateway URI (service stays `dev.arcbox.fleet-agent`), so credentials for different gateways/environments coexist instead of overwriting each other, and `run` loads the entry for the gateway it is configured against. The file backend keeps its per-`data_dir` scoping.

fleet/arcbox-fleet-agent/src/config.rs
PeronGH: Fixed in 8dab0a4b. Parsing is extracted into `parse_load_ceiling`, which requires `n.is_finite() && n > 0.0` — `NaN`, `inf`, zero, and negatives all fail startup now — with a unit test covering the rejects.

fleet/arcbox-fleet-agent/src/runner.rs
PeronGH: Fixed in acc509de. Cancellation is now a level-triggered `CancellationToken` observed at every stage: startup races `docker.start()` against it, and a cancel mid-pull/create/start drops the start future and then awaits a force-remove of the job's deterministically-named container, so a partially-created container is cleaned up and no `RunnerAccepted` is sent. A cancel that lands between stages is still caught by the post-start select because the token is level-triggered rather than a one-shot event.

fleet/arcbox-fleet-agent/src/runner.rs
PeronGH: Fixed in 7850434b, in both redelivery modes. An exact-token replay (worker crash before the dispatch step journaled) is not answered again — the stored verdict stands and the resend loop delivers it. A re-offer under a new token while the accept is still unacked is classified `Duplicate` and re-accepted echoing the new token: replaying the stored verdict verbatim would resolve the old, dead awakeable, so "started here" is re-asserted against the new offer's awakeable instead. No separate accepted-job record was needed — `outstanding` already holds exactly the accepted-but-unsettled window, and the entry is garbage-collected by the gateway's ack.

fleet/arcbox-fleet-agent/src/runner.rs
PeronGH: Fixed in 0b0fff78, with the explicit gateway signal you asked for: arcboxlabs/platform#102 teaches the gateway to distinguish a transient resolve failure (Restate unreachable — no ack, agent keeps resending) from a permanently-gone awakeable (4xx from the ingress — the offer is obsolete), and to ack in the obsolete case too. `OfferVerdictAck` therefore now means "settled: stop resending", and the agent drops `VERDICT_MAX_RESENDS` entirely — accepted verdicts are never expired locally; they are held until the gateway settles them. Growth stays bounded because offers originate from the same parked workflows the acks come from. No wire change (the ack message is unchanged), so no protocol version bump.

fleet/arcbox-fleet-agent/src/runner.rs
PeronGH: Fixed in 0b0fff78. `send_verdict` is now record-then-`try_send`: the `outstanding` entry is the source of truth and the resend loop / next attach delivers on a full or detached queue, so `accept`/`reject` are synchronous and runner supervision never blocks on verdict transport (it stays responsive to cancellation and reaping throughout). As a bonus `handle_provision` and the attach-loop dispatch became fully synchronous, which also makes the check/insert admission window trivially race-free.

AprilNEA (approved):

PR #340, opened 2026-06-25
feat(agent): run under busybox init as a supervised child

AprilNEA:
## Context

Pairs with **arcboxlabs/boot-assets#27** (`feat(rootfs): supervise the guest agent with busybox init as PID 1`). That PR makes busybox `init` PID 1 and `respawn`s the agent; this PR is the agent side of the contract.

Structural problem: the agent was PID 1 (the busybox trampoline `exec`'d it), so any agent crash / `Err` killed PID 1 and panicked the kernel — turning a transient fault into a full VM death. With busybox init as PID 1, the agent becomes a supervised child that is restarted in-VM if it exits.

## Change

- **`main.rs`** — add an `init`/`serve` subcommand split (`parse_mode` → `Mode`). `arcbox-agent init` runs the one-shot system initialization (`init_system`: writable mounts, networking, `/etc`) and exits — this is what rcS invokes during busybox init's sysinit phase. The default / `serve` path runs the long-running vsock agent (respawned by busybox init). The existing `if is_pid1 { … }` block is kept so a **legacy standalone boot** (agent run directly as PID 1, e.g. the e2e harness) still self-initializes; under busybox init the agent is not PID 1, so it skips both `init_system` (rcS already ran it) and the SIGCHLD reaper (PID 1 reaps natively).

- **`shutdown.rs`** — branch `poweroff` on PID. As a supervised child the agent asks PID 1 to power off (`kill(1, SIGUSR2)`) so busybox init stops all children and syncs, with a direct `reboot(POWER_OFF)` fallback (the agent holds `CAP_SYS_BOOT` regardless of PID). The legacy PID-1 path (`kill -1`, sync, reboot) is unchanged.

## Scope

Fixes the **crash/exit** failure mode (agent death no longer panics the kernel). Does **not** address a hung-but-alive agent (host readiness timeout still reboots the VM) or a virtiofs-mount failure (agent + runtime bins both live on virtiofs) — both explicitly out of scope.

## Tests

`cargo test -p arcbox-agent` — +6 unit tests green (`parse_mode` init/serve dispatch; `shutdown_strategy` PID branch). `cargo fmt` + `cargo clippy` clean on host and the aarch64-unknown-linux-musl target (only pre-existing warnings in untouched files remain).

## Coupling / release

Ships in lockstep with boot-assets#27 (the `init` subcommand contract). The `assets.lock` `[boot] version` + `manifest_sha256` bump is a release-time step once the new rootfs is published. End-to-end validation runs via the desktop product / CI (the OSS dev daemon can't mount virtiofs). **Supervision proof:** kill the agent in-guest → busybox init respawns it within ~1s, no kernel panic, no VM reboot.


AprilNEA:
Addressed review feedback in `b2a98ba`:

- **Codex P1 — gate PL011 earlycon to HV backend**: extracted `ensure_earlycon(cmdline, backend)`; only `Hv` gets the pinned `earlycon=pl011,…` directive, `Vz`/`Auto` pass through untouched, explicit operator `earlycon=` respected. Unit tests added.
- **Pullfrog — readiness retry masks genuine guest errors**: both wait loops now remember the last `watch_readiness` error and fold it into the surfaced timeout via `agent_timeout_error` (no longer debug-only).
- **Greptile P2 — stale `remaining` after reconnect**: the async loop recomputes the budget after the reconnect, before `watch_readiness`.
- **Greptile P2 — `grace` semantics**: `shutdown.rs` `poweroff` docstring now explains `grace` under `SignalInit` must cover busybox init's full teardown.
- **Greptile P2 — `arcbox-agent init` exits 0 on failure**: deferred (needs `init_system()` → `Result`) as #341.
- Pulled in Pullfrog's flaky-test fix `bd7b469`.

Paired boot-assets fix for the rcS `timeout` P1: arcboxlabs/boot-assets#27 (`3238a63`).

AprilNEA:
Added E2E coverage for the supervision guarantee (`9eaeab9`):

- **Test-only `KillAgent` RPC** (MessageType `0x000E`/`0x100E`, reusing the `Empty` payload — no protobuf change): host asks the agent to exit so PID 1 (busybox init) respawns it. Follows the `MmapReadFile` test-RPC precedent.
- **`hv_e2e` supervision phase**: boot → ping → kill agent → assert a fresh ping recovers within 30s. Under the old agent-as-PID-1 model this panics the kernel and the VM dies, so a recovered ping is the proof.

`hv_e2e` is a manual/local example (needs HV + Developer ID signing + the busybox-init rootfs from boot-assets#27); it is not run in CI, which has no Apple Silicon VM runner. The boot-assets behaviors also gained a real tree-assembly test (boot-assets#27, `b852a81`).

PeronGH (dismissed):

AprilNEA:
Addressed the remaining Greptile P2 (`arcbox-agent init` exits 0 on critical-mount failure) in `3e891d6`: the one-shot init now re-checks the critical writable layers (/etc, /run, /var, /tmp) and exits non-zero with a specific error instead of silently letting a broken agent respawn. The complementary boot-assets rcS change (act on the exit vs the current `|| true`) is narrowed into #341.

AprilNEA:
**Fail-fast pair complete + boot assets re-pinned to 0.6.1.**

- boot-assets#29 (rcS powers off on a non-zero `arcbox-agent init` instead of respawning a broken agent) is merged and released as **0.6.1** (tag `v0.6.1`, published to the CDN).
- `assets.lock` re-pinned 0.6.0 → 0.6.1 (`9fd107b`), `manifest_sha256 = 6945c6c6…` (verified against `{cdn}/asset/v0.6.1/manifest.json`, schema 0, both arches).

So `arcbox-agent init` now exits non-zero on a critical writable-mount failure (3e891d6) **and** the 0.6.1 rcS acts on it (clean host-driven retry). #341 is fully closed.

PR #346, opened 2026-06-30
Single switchable System VM backend (HV/VZ); remove UtilityVmRole

AprilNEA:
## Summary

Post-FEX, ArcBox runs a **single System VM** (the internal VM that hosts Docker/Sandbox/K8s) whose hypervisor backend is a first-class, persisted, user-switchable choice — no longer inferred from workload "intent". This PR finishes that pivot end to end and removes the now-dead role-routing machinery.

- `VmBackend` collapses to a direct `{ Hv, Vz }` choice (drops `Auto` / `ResolvedBackend` / `resolve_backend`).
- The two utility VMs (Native + Rosetta) collapse into **one System VM**, defaulting to **VZ** (Apple-managed, stable — sidesteps the HV cold-boot agent race). HV (custom VMM) is opt-in.
- x86_64 translation **follows the backend: VZ → Apple Rosetta, HV → FEX.** Either backend can run `linux/amd64`, so there is no separate VM per arch.
- A gRPC RPC (and `abctl system backend`) **switches the backend live**: the choice is persisted in the machine record, and the System VM is restarted on the new backend. The dockerd data image is preserved (containers and images survive), but because the kernel command line differs between backends (HV pins `earlycon=pl011`) the restart detects config drift and recreates the machine record, regenerating SSH host keys.
- `UtilityVmRole` (Native/Rosetta), which was threaded through the whole Docker proxy but always resolved to `Native`, is **removed entirely**, along with the workload registry that only existed to support role routing.

Net: **+732 / −1827** across 43 files.

### Backend selection (`arcbox-vmm`, `arcbox-core`)
- `VmBackend` is a two-variant direct selector the VMM dispatches on straight from `config.backend`; `#[default]` is `Vz`.
- One System VM slot; removed the Rosetta slot, `ROSETTA_*` consts, and `rosetta_role_supported`.
- `enable_rosetta` is wired only on VZ (`rosetta_capable && backend == Vz`); `amd64_runtime_supported` is backend-aware (VZ → Apple-Silicon Rosetta, HV → the `<data_dir>/runtime/bin/FEX` binary), fail-closed.

### Persistence + live switch (`arcbox-core`)
- Backend is persisted in `PersistedMachine.backend` (the machine record), not a side file. This fixes a bug where a persisted HV choice was lost on daemon restart because the machine was reconstructed with the default backend.
- The `Vmm` is built lazily from `VmConfig` at start: `VmManager::set_backend` updates a stopped VM's config (committing the durable record last, so a failed switch can't leave a backend on disk the running system never applied), and `Runtime::switch_system_vm_backend` does graceful-stop → set backend → restart. On restart the differing kernel cmdline trips drift detection, so the machine record is recreated on the new backend (data image preserved, SSH host keys regenerated).

### Control surface (`arcbox-protocol`, `arcbox-api`, `arcbox-cli`)
- `SystemService.GetSystemVmBackend` / `SetSystemVmBackend`.
- `abctl system backend [hv|vz]` — print or switch the backend (switching restarts the System VM).

### Remove `UtilityVmRole` (`arcbox-core`, `arcbox-docker`)
- Replaced the role-keyed runtime API (`ensure_role_ready`, `machine_name_for_role`, `guest_docker_vsock_port_for_role`, `role_is_distinct`, `slot_for`, `role_slots`) with single-System-VM accessors. Deleted `workload.rs`.
- De-parameterized the Docker proxy: dropped the `role` parameter everywhere, collapsed every `*_for_role` proxy variant into its base, removed `GuestConnector::connect_for`, collapsed the per-role readiness tracker to a single `EndpointReadiness`, and deleted `request_context.rs` and `role_resolution.rs` (role resolution always yielded `Native`).
- Removed the workload registry (it became write-only once role resolution was gone); `/containers/{id}/exec` now falls through to the generic proxy fallback.

This is behavior-preserving: no code path ever produced `Rosetta`.

## Verification

- `cargo build` (real binaries) + `make sign-daemon` (Developer ID) — clean.
- `cargo clippy --workspace --all-targets` — zero warnings.
- `cargo fmt --all` + full workspace `--lib` test suite — green (includes a regression test that the backend round-trips through the machine record).

**Not verified live:** the actual VZ↔HV VM reboot needs a signed daemon + the privileged helper + a real VM boot (the HV cold-boot path is known-flaky), which can't run in CI. Smoke test on a Mac:

```
abctl system backend          # -> vz (default)
abctl system backend hv       # switch -> restarts the System VM on HV
abctl system backend          # -> hv ; data + machine record preserved
abctl system backend vz       # switch back
```


AprilNEA:
Rebased onto master (resolved the conflict with the VMM/vm/vm_lifecycle submodule split) and addressed the review findings in `08212ea7`. Verified with the repo toolchain (rustc 1.96 via devenv): `cargo clippy --workspace --all-targets` clean, full `--lib` test suite green.

**Fixed:**

- **Rosetta flag stale after switch** (@pullfrog's main point; greptile *Rosetta Flag Survives HV Switch*; codex *Gate VZ amd64 on Rosetta*). The Rosetta share is now gated by backend in `VmManager::build_vmm_config` — read at every VM build — instead of baked in at create time. `VmConfig.rosetta` is purely the host capability; `enable_rosetta = config.rosetta && backend == Vz`. A stopped-VM `set_backend` therefore recomputes it correctly (no incidental recreate needed). New test `build_vmm_config_gates_rosetta_on_vz_backend` covers VZ→HV→VZ.
- **amd64 admission too loose** (codex). `amd64_runtime_supported` on VZ now also checks `default_vm.rosetta`, so amd64 fails closed when Rosetta is disabled rather than passing admission and failing in the guest.
- **Shutdown failure changes backend** (greptile P1). `switch_system_vm_backend` now propagates a `shutdown` failure (it's a no-op when stopped and force-stops as a fallback, so it only errors when the VM truly can't be torn down) — no backend is recorded for a VM that didn't stop.
- **Backend state splits sources** (greptile P1). `MachineManager::set_backend` commits the persisted record first, then the VM config, then the in-memory record, so a mid-update failure can't leave the three views disagreeing.

**Follow-up (not in this PR):**

- **Proxy readiness stays verified after restart** (greptile P1). This is pre-existing behavior for *any* System VM restart (recovery, drift-recreate, etc.), not new to the switch: the Docker proxy invalidates its endpoint on the next request error and recovers, so a switch costs at most one failed request (after a switch that already restarts the VM and stops all containers). A clean fix is event-driven invalidation (subscribe the proxy to `MachineStopped`/`MachineStarted`); I'll do that as a separate change so it covers all restarts, not just switches.

AprilNEA:
Follow-up: finding **E** (Docker proxy readiness not invalidated on VM restart) is now fixed in `b0d321a2`.

The proxy now subscribes to the runtime event bus and drops its cached guest-dockerd readiness on the System VM's `MachineStopped`. After a backend switch (or any System VM restart) the next Docker request re-verifies against the fresh VM, instead of the previous behaviour where the first request failed on a stale pooled connection and only the second one self-healed.

- `watch_endpoint_invalidation` filters on the System VM machine name, so user-machine stops don't touch the proxy endpoint; `Lagged`/`Closed` are handled, and dropped events still converge via the existing reactive invalidate-on-transport-failure path.
- The server builds `ProxyState` itself (new `router_with_proxy`) so it can hand a clone to the watcher; `create_router` keeps its signature for the integration tests.
- Unit test `invalidates_only_when_system_vm_stops` covers the name filter. clippy + fmt clean, `cargo test -p arcbox-docker` green (verified on rustc 1.96 via devenv).

AprilNEA:
Addressed the latest review round (commits `e40d0cc`, `0f6ba4a`):

**greptile — `server.rs` / `runtime.rs:299` "Stale Vsock Pool Survives Restart" (P1):** Fixed. Invalidating readiness alone wasn't enough — the pooled guest HTTP client kept idle vsock sessions to the stopped VM, so the re-verification `_ping` could still reuse a dead one. Added `GuestHttpClient::reset` (rebuilds the inner hyper client, dropping the pool) and `ProxyState::reset_endpoint` (readiness invalidate + pool drain). The System VM restart watcher now calls `reset_endpoint`; the reactive single-failure path keeps `invalidate_endpoint` (readiness only) so a transient error doesn't flush the whole pool. The switch path reaches this via the `MachineStopped` that `lifecycle.shutdown()` publishes inside `switch_system_vm_backend`.

**greptile — `machine.rs:742` "Persistence Updates First" (P1):** Fixed. `set_backend` now updates the runtime views first and commits the persisted record **last**. Since the lifecycle seeds its backend from that record on restart, committing it only after the in-memory updates succeed means a failed VM-manager update leaves the on-disk backend untouched, so a failed switch can't resurrect the new backend on the next boot.

**pullfrog — "in place / SSH keys kept" wording (ℹ️):** Fixed. Confirmed against the code: an HV↔VZ switch changes the kernel cmdline (`ensure_earlycon` pins `earlycon=pl011` on HV only) → `machine_drift_reason` → `remove(name, true)` → recreate via `create_default_machine`. So the machine record is recreated and SSH host keys regenerated; only the `docker.img` data image (separate persistent path) survives. Corrected the `switch_system_vm_backend` + `set_backend` docs, the inline comment, and the PR body to say so.

clippy + fmt clean, `cargo test -p arcbox-docker -p arcbox-core` green (rustc 1.96 via devenv; 90 + 60 lib tests, 8 proxy integration tests).

AprilNEA:
**greptile — "Proxy Reset Race" (P1):** Fixed in `f1a271b`. You're right that the event watcher raced the restart — `switch_system_vm_backend` could stop+restart before the watcher processed `MachineStopped`.

Replaced the async watcher with a **pull-based check that's synchronous with each request**, so there's no event to lose or race:

- `VmLifecycleManager` bumps a `restart_generation` counter on every stop (inside `shutdown`/`force_stop`, at the point the VM goes down), exposed as `Runtime::system_vm_restart_generation()`.
- The proxy reads it in `ensure_endpoint_verified` (which every request funnels through) and, when it changed since the last verification, invalidates the cached readiness **and** drains the connection pool before re-verifying.

Because the counter is bumped synchronously inside `shutdown` — before `shutdown().await` returns and well before the restart — any request arriving after the stop observes the new incarnation and re-verifies against the fresh VM (blocking on `ensure_ready` if needed) instead of serving a stale `Verified` or cloning a dead pooled client. The only residual window is a request already in flight to the VM at the instant it's torn down, which fails reactively (unavoidable, and pre-existing). The event watcher and `router_with_proxy` are removed.

Unit test `readiness_generation_change_invalidates_verified_state` covers the incarnation-change → invalidate path. Workspace clippy + fmt clean; `cargo test -p arcbox-docker -p arcbox-core` green (rustc 1.96 via devenv).

AprilNEA:
**greptile — "Backend Commit Can Stick" (P1):** Fixed in `3307384`. Correct: if `ensure_ready()` failed after the backend was recorded (e.g. an HV boot hitting the cold-boot agent race), the lifecycle + persisted config already reported the requested backend while the VM was down, and a retry of `abctl system backend <same>` short-circuited on the early same-backend check — returning Ok without starting anything.

`set_backend` has to precede `ensure_ready` (the drift-driven recreate reads the lifecycle backend to rebuild the machine), so I took your second suggested option: the same-backend path now calls `ensure_ready().await?` before returning instead of bailing early. It's a cheap no-op when the VM is already up, and a real (re)start when a prior switch left it down — so a caller can no longer see a successful switch while Docker is unreachable. Updated the doc accordingly.

clippy + fmt clean, `cargo test -p arcbox-core` green (rustc 1.96 via devenv).

PR #347, opened 2026-06-30
chore(assets): fix 0.6.1 manifest SHA to match published build

AprilNEA:
Master pins the 0.6.1 boot manifest SHA as `6945c6c6…`, but the published 0.6.1 manifest (the rcS-init-fail-fast rebuild) is `20c6a934…`. As a result, **any source build off master fails boot-asset verification and cannot start a VM** (`manifest SHA256 mismatch: expected 6945…, got 20c6…`). Installed Homebrew binaries are unaffected (they carry their own matching pin).

Re-pin `manifest_sha256` to the published `20c6…`. Verified: with this pin a freshly built + Developer-ID-signed daemon boots the System VM to ready and serves the Docker API (dockerd 27.5.1, kernel `6.12.11-arcbox`).

One-line, no code changes. Unblocks #346 (and any other source build) once merged.

### DIFF
--- a/assets.lock
+++ b/assets.lock
@@ -13,7 +13,7 @@
 [boot]
 version = "0.6.1"
 cdn = "https://boot.arcboxcdn.com"
-manifest_sha256 = "6945c6c612c7e021ee709f7e50aeedc39ad3cefaf53c6a30b0aaac1919ac49d5"
+manifest_sha256 = "20c6a934cbba0d2b304efdc99af02eae2b9da72a9bfe2c825fb209023d4bd8d7"
 
 [[tools]]
 name = "docker"


AprilNEA:
@pullfrog is correct — closing this. I independently re-verified the live CDN:

```
$ curl -fsSL https://boot.arcboxcdn.com/asset/v0.6.1/manifest.json | shasum -a 256
6945c6c612c7e021ee709f7e50aeedc39ad3cefaf53c6a30b0aaac1919ac49d5  (5432 bytes)
```

The CDN serves `6945…` — exactly what master already pins. master is correct; this PR was inverted.

Root cause of the boot failure that motivated this PR was a **stale local cache**: `~/.arcbox/boot/0.6.1/manifest.json` on the test machine hashes to `20c6a934…` (a transient/reverted rebuild), so boot verification failed against the correct `6945…` pin. The fix is to clear the stale cache and let the daemon re-download, not to change the pin. No change to assets.lock is needed.

PR #349, opened 2026-07-01
fix(docker): route /containers/json + host-teardown correctness + CI routing tests

AprilNEA:
Docker API compat correctness: fixes the routing shadow that broke the desktop app, closes the host-networking teardown gaps found in review, and adds the router-level test coverage that was missing (the whole API surface was untested in CI).

## 1. Routing: `GET /containers/json` 405 (the reported bug)

axum's `DELETE /containers/{id}` route captures the single-segment static endpoints `GET /containers/json` (list) and `POST /containers/prune`, and a bare `fallback` only catches unmatched *paths*, not method mismatches — so those returned `405 Method Not Allowed, Allow: DELETE` instead of proxying. The desktop app's container list hit this. Pre-existing since v0.1.0's route de-enumeration; surfaced now because the 0.4.14 app lists containers via its OpenAPI client.

Fix: a `MethodRouter::fallback` scoped to `/containers/{id}` — a non-DELETE method there proxies; `DELETE /containers/{id}` (incl. a container literally named "json"/"prune") still removes locally. Scoped to this one route so lifecycle routes (create/start/…) still reject wrong methods locally rather than proxying them.

## 2. `kill` teardown on non-fatal signals (review Finding 1)

`kill_container` tore down host port-forwarding + DNS on any Docker kill 204. But Docker returns 204 for *any* delivered signal — `docker kill -s HUP nginx` (reload) would strip a still-running container's networking. Now gated on `kill_terminates_container` (default/SIGKILL only).

## 3. Host networking leaks for API-less stops (review Findings 2 & 3, #352)

Host teardown was 100% driven by the stop/kill/remove handlers, so a container that stops another way — natural exit, `--rm`, `docker prune`, a non-fatal kill that still ends it, OOM, guest-side `docker stop` — leaked its host port-forward rules + DNS. Add a periodic reconciler that lists the guest's running containers and tears down host state for any registered container no longer running. Fail-safe: a guest query error (or nothing registered) skips teardown, so a transient hiccup can't strip live containers. Handler teardown stays as the immediate fast path.

## 4. CI routing coverage (the reason this went unnoticed)

The whole `api_*.rs` integration suite is `#[ignore]`d (needs a running daemon + guest), and CI runs `cargo test --workspace` without `--include-ignored`, so the router's method/path dispatch was never verified — the `GET /containers/json` shadow had an `#[ignore]`d test asserting `200 OK` that could never have passed. `tests/api_routing.rs` builds the real `create_router` with an un-initialised runtime (routing is decided before any handler runs, no guest needed) and asserts the proxied surface isn't 405-shadowed, named-collection deletes stay local, versioned paths route after prefix strip, and wrong methods on lifecycle routes stay local 405s. Plus unit tests for the kill-signal gate and the reconcile decision (orphan teardown / fail-safe / no-op).

## Verification

clippy + fmt clean; `cargo test -p arcbox-docker -p arcbox-core` green on rustc 1.96 (devenv). All new tests run offline in CI.

**Needs a live check on macOS (can't be exercised in CI):** the reconciler's real vsock query to guest dockerd and its 30s cadence — the decision logic and JSON parsing are unit-tested, but the transport + timing want a run against a real guest before relying on it. Everything else is fully verified offline.


AprilNEA:
Addressed the lifecycle-hook concern (greptile P2 / pullfrog) in `7ddb405b`.

You were right that a global `method_not_allowed_fallback` was too broad — it proxied method-mismatches on **every** local route (create, start, stop, …), needlessly widening the proxy surface. The literal create/start bypass isn't reachable (guest dockerd 405s a wrong method, so nothing happens), but relying on that is fragile.

**Fix reworked to scope the fallback to just the one route that needs it.** The shadow only exists on `/containers/{id}` (its `{id}` captures `json`/`prune`), so the method-fallback now lives *there* via `MethodRouter::fallback`:

```rust
.route(
    "/containers/{id}",
    delete(handlers::remove_container).fallback(proxy::proxy_fallback),
)
```

Result:
- `GET /containers/json`, `POST /containers/prune` → proxied (fixed).
- `DELETE /containers/json` / `/prune` (a container literally named that) → still routes to the local remove handler + its host teardown.
- `GET /containers/{id}/start` and any other wrong method on a **lifecycle** route → **still 405 locally**, never proxied. The lifecycle hooks can't be skipped.

Tests now assert all four, including that a wrong method on a lifecycle route stays 405.

Remaining known gap (pullfrog, unchanged by this PR): `POST /containers/prune` removes containers and proxies straight through, so host-side port-forward/DNS teardown doesn't run for prune-removed containers — same as the existing `--rm` guest-side gap. Not a regression (prune 405'd before), and the right fix is host↔guest state reconciliation, which also closes `--rm`. I'll file that as a follow-up issue rather than special-casing prune here.

AprilNEA:
Appended a 10-commit improvement series from a deep review of the docker-proxy layer (as discussed), then ran a multi-agent adversarial review over the whole series and fixed everything it confirmed.

## The series

**Real fixes**
- `929b1577` **DNS staleness on `network connect/disconnect`** — these change a running container's IP/aliases but proxied straight through; now handled locally with a post-op DNS refresh (drop-first, so a container disconnected from its last network doesn't keep a stale address).
- `ff6ceede` **connect-semaphore hardening** — the permit now rides inside the blocking closure: a wedged guest can no longer accumulate unbounded stuck connect threads (cap = 8). Also: monotonic restart-generation via `fetch_max` (stale readers can't regress it), documented the bounded non-101 upgrade read.
- `eba9590b` **error classification** — guest transport failures are now `GuestUnavailable` (502) and only those invalidate the cached readiness; client-side errors no longer cost the next request a re-verification.

**Performance**
- `43fa9a1c` **registry-based teardown resolution** — stop/kill/rm/rename no longer pay a guest inspect round-trip; a host-side name→ID alias registry (exact / alias / unique short-ID prefix) resolves locally. Aliases are registered at start/restart/rename/recovery and reclaimed with the rest of the container's host state.

**Structure**
- `a7fe6387` proxy transport state → `proxy/state.rs`; dead empty route groups deleted.
- `7b4b2847` reconciler shares the router's `ProxyState` pool (one pool, one restart-generation reset).
- `3e6bf888` `guest_query` consolidates the 3 hand-rolled guest queries (handlers / reconciler / daemon recovery); the now-caller-less direct forward variants are deleted.

**CI-real tests**
- `44b7afc5` + `48518e45` mock guest gains a canned-route table; `api_handlers_mock.rs` runs the REAL `create_router` end-to-end offline — readiness, proxying, and host side-effects: start→stop DNS/alias registration+teardown, `kill -s HUP` **not** tearing down networking, versioned DELETE-by-name running local teardown, guest 404 pass-through. The handler path is no longer CI-invisible.

## Adversarial review (22 agents, 4 lenses, 3-verifier majority per finding)

6 findings confirmed, all fixed in `4db34935`:
1. **Alias leak** (medium): alias-only containers (`--network none`, no ports) were invisible to the reconciler → every ephemeral `--rm` run leaked an alias forever. `registered_container_ids` now includes them; reconciler reclaims.
2. **Rename data loss** (medium): a transiently failed post-rename inspect dropped DNS+alias permanently and broke registry teardown for the new name. Now: replace-only-with-usable-data; on failure keep old state + register the new name from the query.
3. **Reset race** (low): pool is reset **before** readiness flips to Unverified, so a concurrent verifier can't `_ping` through a connection to the dead VM.
4. **Test-harness fidelity** (high): the mock harness skipped the production version-strip layer, so versioned requests silently bypassed local handlers in tests; fixed + pinned behaviorally.
5-6. Overclaiming routing assertion fixed; duplicate of (1) merged.

Workspace clippy zero warnings, fmt clean, full workspace tests green (the only failures in a full parallel run are the pre-existing `arcbox-virtio-console` PTY flakes, untouched by this PR and passing in isolation).

PR #350, opened 2026-07-01
fix(hv): deliver virtio-mmio/PL011 SPIs as level-triggered, not edge

PeronGH:
## Summary

Fixes a root cause of the flaky HV (Hypervisor.framework) cold-boot stall (ABX-386): **virtio-mmio / PL011 SPIs were delivered edge-triggered while the guest services them as level-triggered.**

The custom HV path assigns virtio-mmio device INTIDs as `VIRTIO_IRQ_BASE(48) + index` (`darwin_hv/mod.rs`) and never registers them with the `IrqChip` (whose `next_irq` starts at 32; only the generic `register_mmio_device` path calls `allocate_level_irq()`). So those INTIDs have no `IrqConfig`, `deliver_irq`/`deassert_irq` fell through to the `None => Edge` default, and each device completion was delivered as a momentary `set_spi(true); set_spi(false)` **edge pulse**. But the HV FDT (`darwin_hv/setup.rs`) declares every virtio SPI `IRQ_TYPE_LEVEL_HIGH`, so the guest configures and services them as **level** interrupts. Edge-on-level leaves the Apple GIC line in a state the guest re-takes endlessly — an interrupt storm (observed on virtio-rng's `InterruptACK`, `0xc003064`) that intermittently wedges early `/sbin/init` before the agent ever logs.

## Fix

Default unconfigured SPIs (INTID ≥ 32) to level-triggered and honor the deassert for them; SGIs/PPIs (< 32) stay edge. Minimal and contained to `irq/chip.rs`.

## Evidence

Root-caused with a deterministic in-process repro (extended `examples/hv_boot_test.rs` to `vcpu_count == host cores` + real rootfs + real agent) and host-side probes of the guest↔GIC path. At the wedge: all request virtqueues drained (`avail == used`), completions delivered and not suppressed, every vCPU blocked in `hv_vcpu_run` (`wfi=0`, ~0% CPU) — i.e. **not** a virtqueue-notification/EVENT_IDX problem, disproving the prior investigation's theory on mechanism.

**Clean A/B at `vcpu_count == host cores`, identical conditions: edge baseline 0/10 → level fix 3/10 boots reach agent readiness.**

## Not fully fixed

Necessary but not sufficient at full vCPU count (3/10; boot is highly timing-sensitive). The residual is a separate lost-wakeup race — WFI/WFE don't trap on this backend so the guest idles inside `hv_vcpu_run`, and only the net-RX worker calls `hv_vcpus_exit`, not the blk/generic device IRQ paths. Pinning it is gated on guest-side visibility (ABX-388). Full analysis in the ABX-386 comment thread.

## Test plan

- `cargo fmt` / `cargo clippy -D warnings` / `cargo test -p arcbox-vmm irq` green
- New regression tests: `unconfigured_spi_defaults_to_level`, `unconfigured_legacy_irq_stays_edge`
- Manual: HV cold boot at `vcpu_count == host cores` (0/10 → 3/10 vs edge baseline)

Refs ABX-386.

### DIFF
--- a/virt/arcbox-vmm/src/irq/chip.rs
+++ b/virt/arcbox-vmm/src/irq/chip.rs
@@ -9,6 +9,27 @@ use super::{
     MAX_IRQS, TriggerMode,
 };
 
+/// First GICv3 SPI INTID. INTIDs 0–31 are SGIs/PPIs; 32+ are SPIs.
+const SPI_BASE: Irq = 32;
+
+/// Default trigger mode for an IRQ with no explicit `IrqConfig`.
+///
+/// On this GICv3/ARM64 platform every SPI we expose (virtio-mmio devices and
+/// the PL011 UART) is declared `IRQ_TYPE_LEVEL_HIGH` in the guest device tree,
+/// so an SPI must be driven as level-triggered: the line is held asserted
+/// until the guest clears the device's `interrupt_status`, and the device
+/// then deasserts it. Treating an SPI as edge (a momentary assert/deassert
+/// pulse) violates that contract — with the Apple GIC it leaves the line in a
+/// state the guest re-takes endlessly, producing an interrupt storm that wedges
+/// early boot (ABX-386). SGIs/PPIs (< 32) remain edge.
+const fn default_trigger_mode(irq: Irq) -> TriggerMode {
+    if irq >= SPI_BASE {
+        TriggerMode::Level
+    } else {
+        TriggerMode::Edge
+    }
+}
+
 /// IRQ chip abstraction.
 ///
 /// Manages interrupt routing, delivery, and coalescing.
@@ -241,7 +262,7 @@ impl IrqChip {
         let config = configs.get(&irq);
         let trigger_mode = match config {
             Some(c) => c.trigger_mode,
-            None => TriggerMode::Edge,
+            None => default_trigger_mode(irq),
         };
         drop(configs);
 
@@ -300,7 +321,7 @@ impl IrqChip {
         let config = configs.get(&irq);
         let (gsi, trigger_mode) = match config {
             Some(c) => (c.gsi, c.trigger_mode),
-            None => (irq % MAX_GSIS, TriggerMode::Edge),
+            None => (irq % MAX_GSIS, default_trigger_mode(irq)),
         };
         drop(configs);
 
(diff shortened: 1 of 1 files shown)


PeronGH:
Thanks — both reviewers are right, and this changes the conclusion. I re-verified against the tree and against my own repro data:

1. **The `None` branch is unreachable for real virtio SPIs.** `register_virtio_device` (`device/mod.rs:728`) calls `allocate_level_irq()` for every virtio device, producing a **Level** `IrqConfig` with `gsi == irq`. The `VIRTIO_IRQ_BASE(48)+index` assignment I cited in the description is `#[cfg(test)]`-only — my mistake. So `deliver_irq`/`deassert_irq` take the `Some(Level)` path, not the `None` default this PR changes.

2. **My own instrumentation confirms it.** During the repro I recorded the GSIs passed to `set_spi`: blk=`gsi 34`, rng=`gsi 35`. Had the `None` fallback been taken, those would be `34 % MAX_GSIS(24) = 10` and `35 % 24 = 11`. They were 34/35 = `gsi == irq`, which only comes from an explicit config. So the fallback was never load-bearing here.

3. **Therefore this PR is a no-op for ABX-386, and the 0/10 → 3/10 A/B was noise** — the `ABX_FORCE_EDGE` toggle I A/B'd only affected the dead `None` branch, so both builds were behaviorally identical for real devices. HV cold boot is extremely timing-sensitive (host log volume flips it), and 0/10 vs 3/10 is within variance.

4. **The `irq % MAX_GSIS` fallback IS a real latent bug** as flagged (collapses SPI INTIDs into 0–23; should be the INTID directly, matching `allocate_level_irq`'s `gsi == irq`). But it's currently dead code, and this PR would make it *worse* (a held level assert on the wrong line instead of a self-clearing edge pulse).

Given the fix doesn't address ABX-386 and only touches an unreachable fallback, I'll **close this PR** rather than land churn on dead code. The real ABX-386 failure (confirmed) is a lost-wakeup: at `vcpu_count == host cores`, all vCPUs block inside `hv_vcpu_run` (`wfi=0`, ~0% CPU) with every request virtqueue drained and completions delivered — i.e. not an interrupt-config problem. Reopening on that; correcting the ABX-386 thread. Apologies for the wrong turn.

PR #361, opened 2026-07-04
fix(vz): fix MTU override selector detection

chrisnc:
VZFileHandleNetworkDeviceAttachment has maximumTransmissionUnit on macOS 13+. Probing VZVirtioNetworkDeviceConfiguration for this selector instead made supported hosts report the selector unavailable and fall back to 1500.

Probe and send setMaximumTransmissionUnit: on the attachment instead, and correct the availability text from macOS 14 to macOS 13.


chrisnc:
See documentation here: https://developer.apple.com/documentation/virtualization/vzfilehandlenetworkdeviceattachment/maximumtransmissionunit?language=objc

Tested locally before and after and confirmed that the VM's eth0 gets an MTU of 4000 with this fix, and 1500 without.

AprilNEA:
LGTM.

PR #364, opened 2026-07-07
fix(vz): dispatch balloon device access onto the VM serial queue

PeronGH:
## Summary

Fixes **ABX-394**: `arcbox-daemon` SIGTRAPs (`EXC_BREAKPOINT`) ~300s after boot whenever the default Linux (Docker) VM has been idle for the idle timeout — losing every running container/VM. This is the default configuration.

`VZVirtioTraditionalMemoryBalloonDevice` mutations must run on the serial dispatch queue the VM was created on; the framework asserts this via `dispatch_assert_queue`. `MemoryBalloonDevice::set_target_memory_size` issued `setTargetVirtualMachineMemorySize:` directly on the caller's thread, so the idle lifecycle actor (a tokio worker) tripped the assert and aborted the process:

```
dispatch_assert_queue_fail                                          (libdispatch — aborts)
-[VZVirtioTraditionalMemoryBalloonDevice setTargetVirtualMachineMemorySize:]
arcbox_vz::device::balloon::MemoryBalloonDevice::set_target_memory_size
...LifecycleActor::on_idle_tick
```

## Fix

Thread the VM's serial queue into `MemoryBalloonDevice` (mirroring how `VirtioSocketDevice` already carries its queue) and route **both** `set_target_memory_size` and `target_memory_size` through `DispatchQueue::sync` — the same mechanism `start`/`stop`/`pause` use. This makes every caller queue-safe rather than patching only the actor's shrink/restore paths.

Note the getter had the same latent off-queue bug: VZ requires *all* device access on the queue, not just mutations, so it is fixed here too. Stale SAFETY comments updated.

## Verification

Verified end-to-end through the real daemon (release build, codesigned, default VZ backend + balloon device), not a synthetic harness:

- Idled 300s → logged `VM entered idle state after 300s of inactivity` → `Balloon shrunk to 128MB for idle VM`
- Guest kernel processed the inflation (`virtio_balloon ... fill_balloon`, "Out of puff!")
- Daemon stayed **alive**, **zero** crash reports in `~/Library/Logs/DiagnosticReports`

Pre-fix, that `Balloon shrunk` call is exactly where the process trapped. Satisfies the ABX-392 §3 idle-reclaim real-machine verification gate.

`cargo fmt` + `cargo clippy` clean.

### DIFF
--- a/virt/arcbox-vz/src/device/balloon.rs
+++ b/virt/arcbox-vz/src/device/balloon.rs
@@ -85,11 +85,14 @@ impl Drop for MemoryBalloonDeviceConfiguration {
 /// Use `set_target_memory_size()` to adjust the balloon.
 pub struct MemoryBalloonDevice {
     inner: *mut AnyObject,
+    /// The VM's serial dispatch queue. Virtualization.framework requires every
+    /// access to the device — reads and mutations — to run on this queue.
+    queue: *mut AnyObject,
 }
 
-// SAFETY: The inner pointer refers to a VZ framework-managed object. Access is synchronized by the framework's internal dispatch queue.
+// SAFETY: The inner pointer refers to a VZ framework-managed object; every access goes through the VM's serial dispatch queue. `queue` is a thread-safe GCD queue pointer.
 unsafe impl Send for MemoryBalloonDevice {}
-// SAFETY: See above — access is synchronized by the framework's dispatch queue.
+// SAFETY: See above — all access is dispatched onto the VM's serial queue.
 unsafe impl Sync for MemoryBalloonDevice {}
 
 impl MemoryBalloonDevice {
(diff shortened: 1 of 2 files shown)


AprilNEA:
Superseded by the idle-balloon structural redesign that already merged to master (#390/#392/#393): master's `MemoryBalloonDevice` already dispatches every accessor onto the VM's serial queue via a non-owned `DispatchQueue` handle, keeps `from_raw` unsafe with the queue-lifetime contract, and additionally wraps the device enumeration in `vm.rs` in `queue.sync` (which this branch lacks). Every hunk of this diff is a strict downgrade of code now on master, so merging would regress the fix. Closing instead of merging — the ABX-394 crash is already fixed on master.

PR #375, opened 2026-07-08
Expose sandbox nested-virt capability and pre-check create on the host

AprilNEA:
Sandboxes need nested virtualization (`/dev/kvm` inside the System VM), available only on the VZ backend with M3+/macOS 15+. The guest already probes `/dev/kvm` at init and rejects sandbox RPCs with `FAILED_PRECONDITION` — but that's a round-trip into the guest, and clients have no way to know the surface is dead before a `Create`.

This surfaces the capability on the host so `arcbox sandbox create` fails immediately with an actionable message, and clients (CLI/Desktop) can grey out sandbox features without booting a microVM.

- New `SystemService.GetSandboxCapability` RPC returning `{ supported, reason, backend }`, computed from the System VM backend plus a static host nested-virt probe.
- `arcbox_hypervisor::host_supports_nested_virt()` — a cheap cross-platform probe (VZ on macOS, KVM module params on Linux x86_64) that doesn't construct a hypervisor. The Linux detection is refactored out of `KvmHypervisor` to back it.
- `arcbox sandbox create` pre-checks and bails with the reason on unsupported hosts/backends; a missing/unavailable RPC falls through to the create (guest agent remains the backstop).

On supported setups behavior is unchanged.

Closes CORE-13.

### DIFF
--- a/app/arcbox-api/src/system.rs
+++ b/app/arcbox-api/src/system.rs
@@ -9,9 +9,9 @@ use std::sync::Arc;
 use arcbox_grpc::SystemService;
 use arcbox_protocol::v1::{
     Empty, ResolveContainerFsRequest, ResolveContainerFsResponse, ResolveImageFsRequest,
-    ResolveImageFsResponse, SetSystemVmBackendRequest, SetupStatus, SystemVmBackend,
-    SystemVmBackendInfo, VcpuDebug, VirtioDebugInfo, VirtioDeviceDebug, VirtioQueueDebug,
-    setup_status,
+    ResolveImageFsResponse, SandboxCapability, SetSystemVmBackendRequest, SetupStatus,
+    SystemVmBackend, SystemVmBackendInfo, VcpuDebug, VirtioDebugInfo, VirtioDeviceDebug,
+    VirtioQueueDebug, setup_status,
 };
 use tokio::sync::watch;
 use tokio_stream::Stream;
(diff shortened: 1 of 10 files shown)


AprilNEA:
Superseded by #565 (`cb5ed3c3`, merged 2026-08-07), which implemented CORE-13 on the post-Connect architecture:

- Host-side fail-fast on `Create`: `app/arcbox-api/src/connect/control.rs` checks `Runtime::sandbox_nested_virt()` before dialing the guest and returns `FAILED_PRECONDITION` carrying `ErrorCode.NESTED_VIRT_UNSUPPORTED` with the actionable reason (`connect/sandbox_errors.rs`).
- Capability exposure: `SandboxService.GetCapabilities` (answered host-side, no guest round-trip) — consumed by the TS/Python/Rust SDKs and asserted by the sandbox e2e; this replaced the `SystemService.GetSandboxCapability` shape proposed here.
- Probe: `app/arcbox-core/src/sandbox_capability.rs` (VZ `isNestedVirtualizationSupported` on macOS, `/sys/module/kvm_{intel,amd}/parameters/nested` on Linux), with the guest-side `/dev/kvm` probe (412 → FAILED_PRECONDITION) as backstop.

This branch predates the Connect migration (daemon no longer serves tonic) and contains hand-edited generated code, so there is nothing left to salvage. Closing.

PR #377, opened 2026-07-08
fix(agent): match dockerd bridge MTU to the primary NIC

AprilNEA:
## Completes the MTU=4000 rollout (#361 / #376)

#361 made VZ actually raise the guest link MTU to 4000; #376 sized the
host socket so those frames don't `EMSGSIZE`. Hardware validation then
surfaced the remaining gap: **single-flow Host→VM to a default-bridge
container stalls to 0 bits/sec.**

### Root cause

Guest `eth0` = MTU 4000, but Docker's default bridge (`docker0`) stays at
1500. The host datapath injects ~4000-byte frames toward the container;
the guest kernel can't forward them onto the 1500-MTU bridge (DF set, no
working PMTUD) and drops them. Only the inject (Host→VM) direction breaks
— VM→Host is guest-side TCP and respects the peer MSS. #376 was necessary
but couldn't fix a mismatch that lives entirely inside the guest.

### Fix

Derive dockerd `--mtu` from `eth0`'s live MTU, so the default bridge
tracks whatever the host negotiated:
- jumbo link (VZ 4000) → bridge 4000, matched;
- link at 1500 (macOS <13, or a future non-jumbo backend) → no override,
  Docker keeps its default.

No hardcoded coupling to the host's 4000 constant — the guest reads the
actual interface, so it stays correct if that value ever changes.

### Validation (VZ, signed release daemon, Apple Silicon)

Rebuilt the agent (musl), redeployed, rebooted the VM:

| Test | Before | After |
|---|---|---|
| Host→VM, **default bridge** | **0 bit/s** (stall) | **2.2 Gbit/s**, 0 retr |
| VM→Host (`-R`) | 3.6 Gbit/s | 3.5 Gbit/s |
| `docker0` MTU | 1500 | 4000 |

Agent log: `dockerd bridge MTU matched to primary NIC mtu=4000`. Default
`bridge` network now reports `com.docker.network.driver.mtu=4000`.

Added a unit test for the parse/gate (`docker_bridge_mtu`). Note: the
agent's linux tests don't run in CI (workspace gate excludes the crate,
no musl test job) — compile-checked for `aarch64-unknown-linux-musl` and
the fix is hardware-validated above.

> Commit is **unsigned** (1Password ssh signing needs an interactive
> prompt unavailable here); squash-merge re-authors anyway.

### DIFF
--- a/guest/arcbox-agent/src/agent/linux/runtime.rs
+++ b/guest/arcbox-agent/src/agent/linux/runtime.rs
@@ -223,6 +223,22 @@ pub(super) async fn ensure_containerd_ready(
     false
 }
 
+/// Reads the primary NIC's MTU so dockerd's default bridge can match the
+/// host-negotiated link MTU.
+fn primary_nic_mtu() -> Option<u32> {
+    docker_bridge_mtu(&std::fs::read_to_string("/sys/class/net/eth0/mtu").ok()?)
+}
+
+/// Parses a `/sys/class/net/*/mtu` value into a dockerd `--mtu` override.
+///
+/// Returns `None` for an unparseable value or the default 1500 (nothing to
+/// override), so links that stay at 1500 keep Docker's own default and only a
+/// jumbo link (e.g. the VZ 4000 MTU) forces the bridge to match.
+fn docker_bridge_mtu(raw: &str) -> Option<u32> {
+    let mtu: u32 = raw.trim().parse().ok()?;
+    (mtu > 1500).then_some(mtu)
+}
+
 async fn ensure_dockerd_ready(runtime_bin_dir: &Path, notes: &mut Vec<String>) {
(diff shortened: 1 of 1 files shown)


AprilNEA:
Superseded by #378 (`fix(net): clamp host→guest TCP segments to the peer's advertised MSS`).

The datapath now bounds each host→guest segment by the peer's advertised MSS, so the fix is network-agnostic and needs no docker MTU change (docker0 stays 1500). That resolves both review points on this PR without the downsides:
- **@chatgpt-codex-connector (vmnet path):** docker stays 1500, so containers advertise MSS 1460 and never emit jumbo frames onto the ~1518 vmnet bridge.
- **@pullfrog (user-defined/compose networks):** MSS-clamp applies to every network; `--mtu` only sized the default bridge. Hardware-validated: user-network Host→VM 0 → 3.02 Gbit/s.

Closing in favor of #378.

PR #425, opened 2026-07-17
fix(net): reassemble guest-originated IPv4 fragments in the classifier (ABX-429)

AprilNEA:
Fixes ABX-429 (raised by Codex review on #412): the guest kernel IP-fragments UDP datagrams larger than its link MTU, and every datapath consumer (`UdpProxy` egress, `InboundRelay` published-port replies, DHCP/DNS handlers) parses one L4 datagram per frame — so oversized guest-originated UDP was silently dropped in both directions. Pre-existing on all paths; #412 made it visible by fixing the opposite direction.

## What

`FrameClassifier` now routes any IPv4 frame with MF set or a nonzero fragment offset through a new `FragmentReassembler` before L4 classification (`common/splicetcp/src/fragment.rs`):

- Fragments correlate by the RFC 791 `(src, dst, protocol, id)` tuple; received byte ranges are tracked and merged; the completed datagram re-enters classification as one frame with the fragment field cleared, total length rewritten, and header checksum recomputed.
- One seam serves every consumer — egress, inbound relay, DHCP/DNS — on both VMM backends and the `tun_proxy` host harness, with no VMM changes.
- **Guest-controlled input hygiene**: bounds verified before every slice; per-datagram size capped at the 16-bit IP total including an option-bearing-header overflow check; table capped at 32 in-flight datagrams (~2 MiB worst case); incomplete entries evicted after 3 s; unaligned non-final fragments, conflicting finals, and beyond-final fragments reject the entry.

## Validation

- Unit: 11 new tests (in-order/out-of-order/interleaved/duplicate reassembly, timeout eviction, entry cap, beyond-final and unaligned and truncated hostile fragments, classifier-level end-to-end incl. fragmented-DNS kind detection). splicetcp suite 49/49; clippy clean.
- End-to-end A/B on real VZ daemons (probe with explicit sent/serving markers, single 3000-byte datagrams via `python:3-alpine`):

  | direction | baseline (HEAD~1) | this fix |
  |---|---|---|
  | egress: container → host, 3000 B | `SENT` but host receives nothing | `SENT` → host `recv 3000` |
  | inbound reply: published UDP port answers 3000 B | `serving` but client gets 0 | `serving` → client gets 3000 |

- `boot_assets` e2e green over the changed classification path (NFS export, image pull, full container lifecycle; one rerun after a registry-side pull timeout, dockerd log confirmed mid-download).

Found during validation and filed separately: **ABX-431** (proxied egress flow wedged 3 h with upstream death never propagated to the guest — unrelated flow-level bug, evidence attached there).

### DIFF
--- a/common/splicetcp/src/classifier.rs
+++ b/common/splicetcp/src/classifier.rs
@@ -23,6 +23,7 @@ use arcbox_datapath::FrameBuf;
 use arcbox_datapath::pool::PacketPool;
 
 use crate::ethernet::{ArpResponder, ETH_HEADER_LEN};
+use crate::fragment::FragmentReassembler;
 
 /// Default Ethernet MTU (excludes Ethernet header).
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


AprilNEA:
Re the informational note (no per-datagram fragment-count cap): verified the bound math — memory is already capped (≤ ~8.2k disjoint ranges × 16 B ≈ 128 KiB per entry, 32 entries), and the per-push cost is a worst-case ~8k-tuple rebuild, i.e. microseconds — lower amplification than simply flooding full-size frames through the datapath. A benign sender never exceeds ~45 transient ranges (one per out-of-order fragment at 1480 B), so a hard cap of 64 ranges per entry is cheap and strictly safe; adding it as a small hardening commit in the upcoming ABX-431 PR (same crate).

PR #469, opened 2026-07-21
fix(balloon): stop idle-shrink oscillation (fail-open backoff + HV free-page-reporting gate)

AprilNEA:
Balloon-subsystem fixes. Commits 1–2 fix the idle-shrink oscillation; commits 3–4 are cross-repo-audit findings in the same subsystem.

## 1. Back off idle shrink after repeated fail-opens (VZ + HV)

On a memory-pressured **host**, balloon inflation stalls and the controller oscillates indefinitely: VM idles → shrink → inflation churns guest pages through exhausted host swap and stalls → guest hits `available_bytes=0` / floods `Out of puff!` → pressure watch times out → **fail-open** (restore full + note activity → `idle→running`) → 5-min idle timeout → repeat. macOS reports **normal** pressure throughout (`kern.memorystatus_vm_pressure_level == 1`), so there is no host signal to gate on.

**Fix:** track consecutive fail-opens; suppress the idle-shrink entry (no probe) within `fail_open_backoff()` — 10 min, doubling to a 60 min cap. A shrink that **holds** clears the streak, so reclaim self-heals when the host recovers. Corrected the over-claimed `"Oscillation is structurally impossible"` doc.

Observed: a genuinely idle 16 GiB System VM (~700 MiB of containers, 14.6 GiB free) on a host at 113/128 GiB + swap full oscillated every 5 minutes for over an hour — 5477 `Out of puff` lines in one daemon log.

## 2. Skip host-driven idle shrink on the HV backend

HV's `arcbox-virtio-balloon` advertises free page reporting + deflate-on-oom and the guest kernel enables reporting (`CONFIG_VIRTIO_BALLOON` `select PAGE_REPORTING`, verified in `arcboxlabs/kernel`), so the HV guest self-reclaims idle memory. `BalloonDeps::guest_self_reclaims()` (true on HV) gates the inflate/pressure-watch descent off; it runs only on VZ.

## 3. MADV_FREE on macOS (audit finding)

`madvise(MADV_DONTNEED)` is a **no-op reclaim on Darwin** — XNU maps `VM_BEHAVIOR_DONTNEED` to a deactivation hint that never releases physical backing, so every HV balloon inflate / free-page-reporting range freed **zero bytes** while returning success. Switch to `MADV_FREE` on macOS (matching libkrun's macOS balloon path). **This is what makes commit 2's "HV self-reclaims via FPR" real rather than a silent no-op** — without it, HV had no working reclaim at all.

## 4. Guard idle_target() against sub-384-MiB machines (audit finding)

`idle_target()` did `clamp(IDLE_BALLOON_FLOOR, full)`; for a VM configured below the 384 MiB floor that is `min > max`, which `std::clamp` asserts on — **crashing the daemon** on the idle transition of any tiny machine. Cap the floor at `full`.

## Tests

`vm_lifecycle::balloon` — 4 new controller tests + `idle_target_below_floor_machine_does_not_panic`; `arcbox-virtio-balloon` unchanged-green. 160 + 16 pass; clippy + fmt clean.

## Notes / deferred

- End-to-end HV verification still awaits the HV blk-capacity bug fix (HV can't run Docker today). Commits 2–3 are grounded in the device feature bits + verified kernel config, and are backend-gated so VZ is unaffected.
- A host-memory-pressure *gate* was rejected: macOS's own pressure signal reads "normal" under this load. Backoff is signal-agnostic.

### DIFF
--- a/app/arcbox-core/src/vm_lifecycle/actor.rs
+++ b/app/arcbox-core/src/vm_lifecycle/actor.rs
@@ -197,6 +197,13 @@ impl BalloonDeps for RealBalloonDeps {
             .set_balloon_target(&info.vm_id, bytes)
     }
 
+    /// HV advertises virtio-balloon free page reporting + deflate-on-oom
+    /// (`virt/arcbox-virtio-balloon`), so its guest self-reclaims; VZ's
+    /// traditional balloon does not.
+    fn guest_self_reclaims(&self) -> bool {
+        matches!(self.shared.backend(), arcbox_vmm::VmBackend::Hv)
+    }
+
     /// Any failure (connect, RPC, timeout) folds to `None`: the policy treats
     /// "unknown" as its own signal — never shrink on it, and fail open while
     /// shrunk, since a reclaim-thrashing guest is exactly one that cannot
--- a/app/arcbox-core/src/vm_lifecycle/balloon/controller.rs
+++ b/app/arcbox-core/src/vm_lifecycle/balloon/controller.rs
@@ -37,6 +37,13 @@
 //! ([`fail_open_backoff`]): after each failed descent the next idle entry is
 //! suppressed for a growing window, and a shrink that actually holds clears
 //! the streak so idle reclaim self-heals once the host recovers.
+//!
+//! This whole host-driven descent runs only on the VZ backend, whose
+//! traditional balloon exposes nothing but a target. The HV backend's
+//! virtio-balloon advertises free page reporting and deflate-on-oom, so its
+//! guest kernel drains idle memory to the host on its own and recovers under
+//! OOM without any target policy — there the controller stays out of the way
+//! (see [`BalloonDeps::guest_self_reclaims`]).
 
 use std::future::Future;
 use std::sync::Arc;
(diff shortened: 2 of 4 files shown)


AprilNEA:
Closing in favour of #642, which keeps the analysis and drops the code.

Where each of the four commits ended up:

1. **Fail-open backoff.** The oscillation account is kept verbatim in
   `balloon/mod.rs` — including that `kern.memorystatus_vm_pressure_level` read
   *normal* throughout, which is why a host-pressure gate was rejected and a
   signal-agnostic backoff was the only measured answer. The code is not kept:
   the descent it guards has no backend to serve (see below), so the backoff
   would guard nothing.
2. **Skip host-driven shrink on HV.** Correct, and stronger than it looked at the
   time. HV's device does advertise `VIRTIO_BALLOON_F_REPORTING` and the guest
   kernel does enable page reporting, so the guest self-reclaims continuously —
   which means the host-driven descent has no target backend *at all*, because VZ
   can never be reclaim-capable either (guest RAM lives in Apple's XPC helper;
   measured 2026-07-29). #642 states that conclusion and says to delete the
   machinery rather than gate it.
3. **`MADV_FREE` on Darwin.** Superseded by the 2026-07-29 calibration, which is
   already in the tree: `MADV_DONTNEED` is footprint-inert as this PR said, but
   the primitive that drops footprint instantly is `MADV_FREE_REUSABLE`, and that
   is what `virt/arcbox-virtio-balloon/src/lib.rs` now names as the remaining HV
   work. Same finding, sharper answer.
4. **`idle_target()` sub-384-MiB guard.** Carried into #642 as a standalone fix.
   `clamp(min > max)` is a std assert inside the lifecycle actor, so it takes the
   daemon down on a tiny machine's first idle transition; that is worth fixing
   before someone flips `reclaim_capable`, not after.

PR #688, opened 2026-08-19
fix(computer): route detach_all through the per-computer actor

PeronGH:
Closes CORE-145.

`SandboxManager::detach_all` was the one terminal transition that did not go
through the per-computer actor. It cloned each `Arc<dyn VmHandle>` straight out
of the map and called the driver, so a graceful-shutdown handover could run
concurrently with the `Stop` or `Remove` of the same computer.

Those are not independent events. An in-flight `StopSandbox` carrying a 30s
budget is the main reason a composer's drain deadline expires and `detach_all`
gets called at all — and dropping a tonic server future does not cancel the
handler, because connections are spawned (`serve_connection` is a bare
`tokio::spawn`; the outer future only awaits `signal_tx.closed()`). So the stop
keeps running while the composer detaches.

## What each order did to the guest

**Stop after detach** is the expensive one, and the likelier interleaving since
the detach is what the timeout triggers:

- `FcProcess::detach` tells the reaper to stand down and `Exited` is never
  published for a detached process, so `stop_vm`'s `handle.shutdown(Graceful)`
  waits out its **entire** budget for an exit that cannot arrive, warns, and
  falls through to `process.kill()` — SIGKILL on the VM that was just reported
  as successfully handed over — then returns `ReapTimeout`.
- `shutdown` also runs `unlink_vsock()`, removing the socket the successor's
  adopted handle dials.
- `release_scope(Runtime)` then hands TAP, the CoW device and the jail back
  while the successor is adopting them.

**Detach during a stop** hands the next process a VM a teardown is halfway
through killing.

## The fix

Detach becomes a lifecycle transition of its own, so the actor serializes it
against the other two.

- `Event::Detach` sits on the `active` superstate — exactly the three states
  holding a settled live handle — beside `Event::Stop`. It emits
  `Effect::Detach` and does **not** transition: the port call decides, and a
  handover that fails must leave a computer that is still ours and still usable
  rather than needing a state to come back from.
- The effect is awaited in the dispatch rather than spawned like every other
  flow. Releasing ownership has no wait in it, and spawning would force an
  intermediate state whose failure arm has nowhere correct to go — `failed`
  would write `Failed` over a record the successor's sweep still has to read as
  `Ready`. `queued` already exists for feeding an effect's outcome back into
  the same dispatch.
- Success lands in a new `detached` state, deliberately **not** under the
  `computer` superstate, so `Remove`, the TTL cap, `VmExited` and every failure
  event are swallowed. That placement is what closes the stop-after-detach path
  above.
- Nothing durable is written. The successor's sweep adopts from a `Ready`
  record; any other phase would have it reinstated as `Failed` instead.

`detach_all` now fans out over the mailboxes, keeps its best-effort
collect-and-report contract, skips an actor that finished in between
(`NotFound` — its computer is already gone), and waits out the startup sweep
first as `stop_sandbox`/`remove_sandbox` do.

Refusals fall out of the existing "machine had nothing to do → the actor
decides what the caller is told" contract. A computer mid-launch or
mid-teardown answers `WrongState`; one with nothing to hand over (paused,
resting, already detached) answers `Ok`. That answer keys off the machine
state, not the runtime handle — a computer halfway through a stop still holds
one, which is exactly the case that must not report success.

## Invariants

Three existing invariants asserted rules the new state breaks by design. Each
was extended to name the exception rather than drop the case, and
`a_handed_over_computer_refuses_every_teardown` pins why:

- a non-forced remove is refused on a handed-over computer,
- a forced remove does not tear one down,
- the TTL cap does not destroy one.

## Tests

- `cargo test -p arcbox-computer-runtime --lib` — 210 pass, incl. new machine
  invariants (`a_handover_only_acts_where_a_live_handle_is_settled`,
  `a_handed_over_computer_refuses_every_teardown`,
  `a_handover_drops_the_timers_it_no_longer_owns`) and actor tests over the
  `ComputerTasks` seam (handover during a stop refused; teardown after a
  handover never reaching the VM; a refused handover leaving the computer
  usable).
- `cargo test -p arcbox-computer-runtime --test manager_over_fakes` — 34 pass.
  New: `a_teardown_after_a_handover_never_reaches_the_vm` (asserts
  `shutdowns(&vm)` is empty and no network cleanup ran) and
  `a_handover_during_a_stop_is_refused_and_reported`. The three pre-existing
  `detach_all` tests are unchanged, which is the compatibility proof.
- `cargo fmt --check`, `cargo clippy` (zero new warnings), `cargo xtask
  check-layers` — clean.

`tests/e2e.rs` also calls `detach_all` but is `#[ignore]`d behind
`FC_BINARY`/`FC_KERNEL`/`FC_ROOTFS` and root, so it needs the `test-vm-linux`
runner rather than this host.

## Expected consumer

`detach_all` has no production caller in this repo — only `tests/manager_over_fakes.rs` and the `#[ignore]`d `tests/e2e.rs`. The consumer is the **platform node-agent** (`platform#146`), which calls it from its graceful-shutdown path once its drain deadline expires; the platform is holding for this fix rather than writing a workaround. So the change lands ahead of its caller by design, and the invariants above are enforced by the HSM and its tests rather than by any path CI exercises end to end.

## A deliberate trade: `Detached → SandboxState::Ready`

A client sees `Ready` for a computer that refuses every lifecycle verb with a wrong-state error. That is the honest option of the ones available — the wire enum has no `Detached` variant, the guest really is running, and the successor adopts it as `Ready` — but it is a trade, not a free mapping, and it is the reason two of the review findings existed: `SetLifecycle` and the data plane both gated on the public projection and therefore let a handed-over computer through. Both now gate on the machine state instead. Anything else that reasons about a handed-over computer should do the same: the state is authoritative, the projection is not.

### DIFF
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
@@ -132,6 +132,42 @@ impl ComputerActor {
                     }
                 }
             }
+            Command::Detach { reply } => {
+                // Parked *before* the dispatch, like a removal and for a
+                // sharper reason: the handover is awaited inside the effect
+                // rather than spawned, so `Answer::Detached` always lands
+                // before this dispatch returns.
+                self.waiters.push((Answer::Detached, reply));
+                if self.dispatch(machine, Event::Detach).await {
+                    return;
+                }
+                // The machine had nothing to do, which is two different
+                // answers — and the state is what tells them apart, not the
+                // handle: a computer halfway through a stop still holds one.
+                let answer = match machine.state() {
+                    // Nothing to hand over, which is a real success: there is
+                    // no guest here this process would have killed on its way
+                    // out. A paused computer gave its VM up to its checkpoint
+                    // and a resting one has none, so the successor reinstates
+                    // all of them from the record alone. An already-detached
+                    // one is the successor's twice over — reporting a failure
+                    // would have a composer log a loss it did not take.
+                    State::Paused {}
+                    | State::Stopped {}
+                    | State::Failed {}
+                    | State::Gone {}
+                    | State::Detached {} => Ok(()),
+                    // Mid-launch or mid-teardown. The launch's resources are
+                    // still in its task rather than on the computer, and a
+                    // teardown was asked for — either way a caller that heard
+                    // `Ok` would believe a guest survived a handover that
+                    // never happened.
+                    _ => Err(self.wrong_state("Ready, Running, or Checkpointing")),
+                };
+                if let Some((_, reply)) = self.waiters.pop() {
+                    let _ = reply.send(answer);
+                }
+            }
             Command::ClaimWorkload { claim, reply } => {
                 let taken = self.dispatch(machine, Event::ClaimWorkload { claim }).await;
                 let _ = reply.send(if taken {
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
@@ -11,6 +11,22 @@ use super::*;
 
 impl ComputerActor {
     pub(super) async fn on_command(&mut self, machine: &mut Machine, command: Command) {
+        // A handed-over computer answers nothing but another handover.
+        //
+        // One rule here rather than a guard per command, because the thing
+        // that makes this state easy to miss is systematic: `detached`
+        // projects `Ready` — the wire has no variant for "the successor's" —
+        // so every gate below that reads `self.public()` is blind to it. The
+        // ones that refuse it today do so by luck of their match arms, and
+        // `Resume` did not: it read `Ready` and answered `Ok` for a computer
+        // this process no longer owned. A rule at the door cannot be
+        // forgotten by whatever command is added next.
+        if matches!(machine.state(), State::Detached {})
+            && !matches!(command, Command::Detach { .. })
+        {
+            self.refuse_handed_over(command);
+            return;
+        }
         match command {
             Command::Provision {
                 provision,
@@ -199,18 +215,6 @@ impl ComputerActor {
                 self.dispatch(machine, Event::WorkloadReleased).await;
             }
             Command::SetLifecycle { update, reply } => {
-                // A handed-over computer refuses too, and the state is what
-                // says so — it projects `Ready`, so the public gate below lets
-                // it through. What that would reach is not a timer but the
-                // record: `persist_lifecycle` fsyncs into the record the
-                // successor has already adopted, under the generation it
-                // adopted with, so the write is accepted rather than fenced.
-                // The same class as the VM race this whole transition exists
-                // to close, moved from the guest to its record.
-                if matches!(machine.state(), State::Detached {}) {
-                    let _ = reply.send(Err(self.wrong_state("a computer this process still owns")));
-                    return;
-                }
                 // `set_sandbox_lifecycle` refuses a computer on its way out:
                 // nothing is left for a deadline to fire on, and the record
                 // it would persist to is about to be a tombstone.
@@ -258,6 +262,40 @@ impl ComputerActor {
                     .get_or_insert_with(|| format!("computer {} exited unexpectedly", self.id));
                 self.dispatch(machine, Event::VmExited).await;
             }
+        }
+    }
+
+    /// Answers every caller of a computer this process has handed over.
+    ///
+    /// What it protects is not one command. `SetLifecycle` was the loud case —
+    /// `persist_lifecycle` fsyncs into the record the successor has already
+    /// adopted, under the generation it adopted with, so the write is accepted
+    /// rather than fenced, which is the VM race this transition closes moved
+    /// onto the record. But `Resume` answered `Ok` for the same reason, and
+    /// the next command would have too.
+    ///
+    /// The match stays exhaustive so that adding one forces a decision here.
+    /// `Detach` cannot reach this — the caller lets it through, because a
+    /// second handover is an idempotent no-op rather than a refusal — and the
+    /// tells have nobody to answer: `detached` swallows them anyway.
+    fn refuse_handed_over(&self, command: Command) {
+        let refused = || self.wrong_state("a computer this process still owns");
+        match command {
+            Command::Provision { reply, .. }
+            | Command::Pause { reply, .. }
+            | Command::Resume { reply, .. }
+            | Command::Stop { reply, .. }
+            | Command::Remove { reply, .. }
+            | Command::ClaimWorkload { reply, .. }
+            | Command::SetLifecycle { reply, .. }
+            | Command::Detach { reply } => {
+                let _ = reply.send(Err(refused()));
+            }
+            // Its own reply type, so it cannot join the arm above.
+            Command::Checkpoint { reply, .. } => {
+                let _ = reply.send(Err(refused()));
+            }
+            Command::WorkloadExited { .. } | Command::ReleaseWorkload | Command::VmExited => {}
         }
     }
 
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
@@ -6,6 +6,7 @@
 //! is also where a caller parks until the effect that answers it lands.
 
 use chrono::Utc;
+use tracing::warn;
 
 use super::*;
 
@@ -149,6 +150,30 @@ impl ComputerActor {
                 }
             }
             Command::Detach { reply } => {
+                // The handover answers on its own outcome and nothing else.
+                //
+                // Answer state that is still set belongs to a flow that ended
+                // with nobody parked to hear it — an idle-driven pause whose
+                // `Pausing` write was refused is the reachable one, since
+                // `Durability::Warn` softens only an unconfirmed write, not a
+                // refused one. `answer` would fold it into this reply and
+                // report a guest lost that was in fact handed over, and
+                // unretryably: the state is terminal by then, so the retry
+                // answers `Ok` and contradicts the first answer. Going through
+                // `park` would report it *instead of* attempting the handover,
+                // which is the same false loss one step earlier. So it is
+                // dropped here, and logged, because nothing else can report it
+                // now. (`Remove` carries the same hazard and predates this;
+                // its answer at least stays consistent under a retry.)
+                let stale_failure = self.answer_error.take();
+                let stale_unconfirmed = self.unconfirmed.take();
+                if let Some(detail) = stale_failure.or(stale_unconfirmed) {
+                    warn!(
+                        sandbox_id = %self.id,
+                        detail,
+                        "dropping an unreported failure from an earlier flow; the handover answers for itself"
+                    );
+                }
                 // Parked *before* the dispatch, like a removal and for a
                 // sharper reason: the handover is awaited inside the effect
                 // rather than spawned, so `Answer::Detached` always lands
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
@@ -148,21 +148,24 @@ impl ComputerActor {
                     // Nothing to hand over, which is a real success: there is
                     // no guest here this process would have killed on its way
                     // out. A paused computer gave its VM up to its checkpoint
-                    // and a resting one has none, so the successor reinstates
-                    // all of them from the record alone. An already-detached
-                    // one is the successor's twice over — reporting a failure
-                    // would have a composer log a loss it did not take.
-                    State::Paused {}
+                    // and a resting one has none; `provisioning` has not
+                    // spawned its boot yet, so no VMM exists either. The
+                    // successor reinstates all of them from the record alone.
+                    // An already-detached one is the successor's twice over —
+                    // reporting a failure would have a composer log a loss it
+                    // did not take.
+                    State::Provisioning {}
+                    | State::Paused {}
                     | State::Stopped {}
                     | State::Failed {}
                     | State::Gone {}
                     | State::Detached {} => Ok(()),
-                    // Mid-launch or mid-teardown. The launch's resources are
-                    // still in its task rather than on the computer, and a
-                    // teardown was asked for — either way a caller that heard
-                    // `Ok` would believe a guest survived a handover that
-                    // never happened.
-                    _ => Err(self.wrong_state("Ready, Running, or Checkpointing")),
+                    // Mid-launch, mid-capture or mid-teardown — all of which
+                    // do have a VM, and all of which lose it when this process
+                    // exits. That is worth reporting rather than skipping: a
+                    // caller told `Ok` would believe a guest survived a
+                    // handover that never happened.
+                    _ => Err(self.wrong_state("Ready or Running")),
                 };
                 if let Some((_, reply)) = self.waiters.pop() {
                     let _ = reply.send(answer);
@@ -196,6 +199,18 @@ impl ComputerActor {
                 self.dispatch(machine, Event::WorkloadReleased).await;
             }
             Command::SetLifecycle { update, reply } => {
+                // A handed-over computer refuses too, and the state is what
+                // says so — it projects `Ready`, so the public gate below lets
+                // it through. What that would reach is not a timer but the
+                // record: `persist_lifecycle` fsyncs into the record the
+                // successor has already adopted, under the generation it
+                // adopted with, so the write is accepted rather than fenced.
+                // The same class as the VM race this whole transition exists
+                // to close, moved from the guest to its record.
+                if matches!(machine.state(), State::Detached {}) {
+                    let _ = reply.send(Err(self.wrong_state("a computer this process still owns")));
+                    return;
+                }
                 // `set_sandbox_lifecycle` refuses a computer on its way out:
                 // nothing is left for a deadline to fire on, and the record
                 // it would persist to is about to be a tombstone.
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
@@ -181,7 +181,18 @@ impl ComputerActor {
                     // exits. That is worth reporting rather than skipping: a
                     // caller told `Ok` would believe a guest survived a
                     // handover that never happened.
-                    _ => Err(self.wrong_state("Ready or Running")),
+                    //
+                    // Phrased as a property rather than a state list, because
+                    // `wrong_state` reads `actual` off the public projection
+                    // and two states here project into any list this arm could
+                    // name: `checkpointing` reads `Ready` and a `gating` whose
+                    // own `cmd` has claimed the slot reads `Running`. Naming
+                    // the states would tell a composer "expected Ready or
+                    // Running, actual Ready" — and `detach_all` folds this
+                    // verbatim into the failure string someone diagnoses a
+                    // lost handover from.
+                    _ => Err(self
+                        .wrong_state("a computer with no launch, capture, or teardown in flight")),
                 };
                 if let Some((_, reply)) = self.waiters.pop() {
                     let _ = reply.send(answer);
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
@@ -124,29 +124,47 @@ impl ComputerActor {
                     }
                 });
             }
-            Effect::Detach => match self.tasks.detach().await {
-                // The machine has not moved yet: it emitted this effect and
-                // stayed put, so the handover's own outcome is what decides.
-                Ok(()) => {
-                    // The data plane reads the agent straight off the snapshot
-                    // and never round-trips the mailbox, so the state alone
-                    // cannot stop it — and `detached` projects `Ready`, which
-                    // is exactly what `require_alive_agent` admits. Dropping
-                    // the agent is what keeps an exec or a file write out of a
-                    // guest the successor now owns; a stop and a release do
-                    // the same for the same reason.
-                    self.forget_agent();
-                    self.queued.push_back(Event::Detached);
+            Effect::Detach => {
+                // The data plane reads the agent straight off the snapshot and
+                // never round-trips the mailbox, so the state alone cannot
+                // stop it — and `detached` projects `Ready`, which is exactly
+                // what `require_alive_agent` admits. Dropping the agent is
+                // what keeps an exec or a file write out of a guest the
+                // successor now owns; a stop and a release do the same.
+                //
+                // It goes *before* the port call, not after. Detach is the one
+                // flow whose ownership transfer happens inside an await, so
+                // forgetting afterwards would leave the whole call open for a
+                // reader to take the agent out of the snapshot and dial a VM
+                // that had already changed hands. This does not close the race
+                // — a reader that cloned the `Arc` a moment earlier still
+                // holds a working agent, and revoking that would mean routing
+                // the data plane through the mailbox, which is the hop this
+                // seam exists to avoid — but it bounds the exposure to callers
+                // already in flight rather than every caller for the duration.
+                let agent = self.snapshot_tx.borrow().agent.clone();
+                self.forget_agent();
+                match self.tasks.detach().await {
+                    // The machine has not moved yet: it emitted this effect
+                    // and stayed put, so the handover's outcome is what
+                    // decides.
+                    Ok(()) => self.queued.push_back(Event::Detached),
+                    // Stay where we are. The VM is still ours and still usable
+                    // — it simply dies with this process, which is what a
+                    // failed handover has always meant. Failing the computer
+                    // instead would write `Failed` over a record the
+                    // successor's sweep still has to read as `Ready`. Usable
+                    // includes dialable, so the agent goes back: nothing
+                    // changed hands, and the data plane must not have lost a
+                    // guest to a handover that did not happen.
+                    Err(failure) => {
+                        if let Some(agent) = agent {
+                            self.publish_agent(agent);
+                        }
+                        self.fail_waiters(failure.into_error());
+                    }
                 }
-                // Stay where we are. The VM is still ours and still usable —
-                // it simply dies with this process, which is what a failed
-                // handover has always meant. Failing the computer instead
-                // would write `Failed` over a record the successor's sweep
-                // still has to read as `Ready`.
-                Err(failure) => {
-                    self.fail_waiters(failure.into_error());
-                }
-            },
+            }
             Effect::AbortInflight => return self.abort_inflight().await,
             Effect::Publish(notify) => self.publish(notify),
             Effect::ArmTimer(timer) => self.arm(timer, state),
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
@@ -127,7 +127,17 @@ impl ComputerActor {
             Effect::Detach => match self.tasks.detach().await {
                 // The machine has not moved yet: it emitted this effect and
                 // stayed put, so the handover's own outcome is what decides.
-                Ok(()) => self.queued.push_back(Event::Detached),
+                Ok(()) => {
+                    // The data plane reads the agent straight off the snapshot
+                    // and never round-trips the mailbox, so the state alone
+                    // cannot stop it — and `detached` projects `Ready`, which
+                    // is exactly what `require_alive_agent` admits. Dropping
+                    // the agent is what keeps an exec or a file write out of a
+                    // guest the successor now owns; a stop and a release do
+                    // the same for the same reason.
+                    self.forget_agent();
+                    self.queued.push_back(Event::Detached);
+                }
                 // Stay where we are. The VM is still ours and still usable —
                 // it simply dies with this process, which is what a failed
                 // handover has always meant. Failing the computer instead
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
@@ -144,7 +144,19 @@ impl ComputerActor {
                 // already in flight rather than every caller for the duration.
                 let agent = self.snapshot_tx.borrow().agent.clone();
                 self.forget_agent();
-                match self.tasks.detach().await {
+                // Bounded because this one is awaited inline; see
+                // [`DETACH_TIMEOUT`]. A driver that hangs here loses this
+                // computer, not the whole pass.
+                let handover = match tokio::time::timeout(DETACH_TIMEOUT, self.tasks.detach()).await
+                {
+                    Ok(handover) => handover,
+                    Err(_) => Err(TaskFailure::recoverable(VmmError::Process(format!(
+                        "handing computer {} over did not finish within {}s",
+                        self.id,
+                        DETACH_TIMEOUT.as_secs()
+                    )))),
+                };
+                match handover {
                     // The machine has not moved yet: it emitted this effect
                     // and stayed put, so the handover's outcome is what
                     // decides.
--- a/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
@@ -124,6 +124,19 @@ impl ComputerActor {
                     }
                 });
             }
+            Effect::Detach => match self.tasks.detach().await {
+                // The machine has not moved yet: it emitted this effect and
+                // stayed put, so the handover's own outcome is what decides.
+                Ok(()) => self.queued.push_back(Event::Detached),
+                // Stay where we are. The VM is still ours and still usable —
+                // it simply dies with this process, which is what a failed
+                // handover has always meant. Failing the computer instead
+                // would write `Failed` over a record the successor's sweep
+                // still has to read as `Ready`.
+                Err(failure) => {
+                    self.fail_waiters(failure.into_error());
+                }
+            },
             Effect::AbortInflight => return self.abort_inflight().await,
             Effect::Publish(notify) => self.publish(notify),
             Effect::ArmTimer(timer) => self.arm(timer, state),
--- a/computer/arcbox-computer-runtime/src/lifecycle/machine.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/machine.rs
@@ -365,10 +365,18 @@ impl ComputerLifecycle {
             // this process no longer owns. Refusing costs the departing
             // process's capture; accepting would corrupt the successor's
             // guest, and the caller hears which computer it lost.
+            // `Detached` is swallowed for the same reason as the `Detach` that
+            // would have produced it, and not only because none can arrive
+            // today: `Effect::Detach` is awaited inline, so its completion is
+            // drained inside the dispatch that asked for it, and this state
+            // never asks. Were it ever spawned like every sibling flow, a
+            // completion landing here would inherit `active`'s arm and detach
+            // mid-capture — exactly the corruption the refusal above prevents.
             Event::ClaimWorkload { .. }
             | Event::Pause { .. }
             | Event::Checkpoint
             | Event::Detach
+            | Event::Detached
             | Event::IdleExpired { .. } => Handled,
             _ => Super,
         }
--- a/computer/arcbox-computer-runtime/src/lifecycle/machine.rs
+++ b/computer/arcbox-computer-runtime/src/lifecycle/machine.rs
@@ -238,8 +238,10 @@ impl ComputerLifecycle {
         }
     }
 
-    /// The three states holding a settled live handle — which is exactly the
-    /// set a handover can act on, and why `Detach` sits beside `Stop` here.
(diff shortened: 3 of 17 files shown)


computer/arcbox-computer-runtime/src/lifecycle/machine.rs
```diff
@@ -238,13 +238,31 @@ impl ComputerLifecycle {
...
                 context.stop(*budget_ms, false);
                 Transition(State::stopping())
             }
+            // No transition: the port call decides. A handover that fails
+            // leaves a computer that is still ours and still usable, so the
+            // machine must not have moved off the state it came from.
+            Event::Detach => {
+                context.emit(Effect::Detach);
+                Handled
+            }
```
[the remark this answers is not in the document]
PeronGH: Confirmed and fixed in ec73eddf — this was a real break of the invariant the PR sets out to establish, and the trace is exactly right: `checkpointing` delegates to `active`, `Effect::Detach` is awaited inline so it never reaches `spawn`, and the arm emitted no `AbortInflight`. `checkpoint_impl` clones `inst.handle` up front and pauses/snapshots/resumes through it, and `FcHandle::detach` only stands the reaper down, so the handle stays fully functional.

`checkpointing` now refuses the handover (added to its existing `Handled` arm list) rather than aborting the capture. Aborting is the worse of the two: a capture cut off mid-write can leave the guest quiesced, and the port has no verb that thaws one — so preempting would hand the successor a frozen guest instead of a mutated one. Refusing costs the departing process its capture, and `detach_all` reports which computer it lost.

Regressions: `a_handover_is_refused_while_a_capture_is_in_flight` (machine-level, also asserts the handover works normally once `CaptureDone` lands) and `a_handover_during_a_capture_is_refused` (actor-level, asserts the port was never reached and the capture still answers its caller).

computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
```diff
@@ -124,6 +124,19 @@ impl ComputerActor {
                     }
                 });
             }
+            Effect::Detach => match self.tasks.detach().await {
+                // The machine has not moved yet: it emitted this effect and
+                // stayed put, so the handover's own outcome is what decides.
+                Ok(()) => self.queued.push_back(Event::Detached),
```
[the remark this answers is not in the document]
PeronGH: Confirmed and fixed in ec73eddf. `Effect::Detach` now calls `forget_agent()` on success, as `SpawnStop` and `SpawnRelease` already did.

The gap was exactly as described: `agent_in` (`sandbox/lifecycle.rs`) reads the agent off the `ComputerSnapshot` with no mailbox round-trip — deliberately, it is the data-plane seam — and admits `Ready`/`Running`, which `Detached → Ready` satisfies. So every exec and file verb still dialed the handed-over guest.

Regression: `a_handover_takes_the_guest_agent_off_the_snapshot`.

computer/arcbox-computer-runtime/src/lifecycle/machine.rs
```diff
@@ -238,13 +238,31 @@ impl ComputerLifecycle {
...
             Event::Stop { budget_ms } => {
                 context.stop(*budget_ms, false);
                 Transition(State::stopping())
             }
+            // No transition: the port call decides. A handover that fails
+            // leaves a computer that is still ours and still usable, so the
+            // machine must not have moved off the state it came from.
+            Event::Detach => {
+                context.emit(Effect::Detach);
+                Handled
```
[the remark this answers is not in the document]
PeronGH: Confirmed and fixed in ec73eddf — same finding as the pullfrog thread on `machine.rs`. `checkpointing` now refuses the handover instead of inheriting the `active` arm; see that thread for why refusing beats preempting the capture.

computer/arcbox-computer-runtime/src/lifecycle/projection.rs
```diff
@@ -28,6 +28,10 @@ impl State {
             Self::Capturing {} | Self::Releasing {} => SandboxState::Pausing,
             Self::Paused {} => SandboxState::Paused,
             Self::Stopping {} | Self::Removing {} => SandboxState::Stopping,
+            // A handed-over computer is still a live, usable guest — it just
+            // belongs to the next process. That is what its record says and
+            // what the successor adopts it as, so it is what a caller sees.
+            Self::Detached {} => SandboxState::Ready,
```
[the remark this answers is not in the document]
PeronGH: Confirmed and fixed in ec73eddf. `SetLifecycle` now refuses on `State::Detached {}` before the public-state gate.

Verified the whole chain, including the part that makes it bite: the guard is `matches!(self.public(), Stopping | Stopped | Failed)`, `Detached` projects `Ready` and sails through, and `persist_lifecycle` → `records.update_lifecycle()` is an fsync'd write. The generation fence does not catch it because `reconcile.rs` adopts with `record.generation` — the same UUID the departing actor still holds — so the write is accepted, not rejected. Agreed on the characterisation: the same class as the VM race, moved to the record.

Gated on `machine.state()` rather than adding a `SandboxState` variant, since the state is the authoritative signal here and the public projection deliberately is not. Also agreed the `rearm` half is harmless; the fix covers both anyway by refusing before either runs.

Regression: `a_handed_over_computer_refuses_a_lifecycle_update`.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
```diff
@@ -543,25 +543,37 @@ impl SandboxManager {
...
+    /// order damaged the guest: a stop reaching a detached handle waits out a
+    /// stood-down reaper and then SIGKILLs the VM it was just told had been
+    /// handed over, and its release hands TAP, the CoW device and the jail back
+    /// while the successor is adopting them.
     pub async fn detach_all(&self) -> Result<()> {
-        let live: Vec<(SandboxId, Arc<dyn VmHandle>)> = self
+        // The sweep adopts and kills VMs by the same deterministic names, so
+        // racing it is the same class of bug; `stop_sandbox` and
+        // `remove_sandbox` wait it out for exactly this reason.
+        self.await_reconcile().await?;
```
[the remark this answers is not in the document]
PeronGH: Two halves, one taken and one refuted with reasoning — thanks for splitting them out, both deserved a deliberate decision rather than falling out of the rewrite.

**The `?` on `await_reconcile`: taken.** Fixed in ec73eddf. The argument is right — `stop_sandbox` can afford to fail one computer, but this is the process-wide last chance, and aborting the whole handover because the sweep failed kills every guest to report a fault in something that has already finished. It now warns and hands over anyway. The wait itself stays: that is what keeps the handover from racing the sweep's adopt/kill by deterministic per-id names, which is the same bug class this PR closes.

**The unfiltered fan-out: mostly keeping it, with one correction.** `provisioning` now answers `Ok` — it has not spawned its boot yet, so there is provably no VMM, and reporting it as lost was simply wrong. That was a genuine bug in my split.

For `staging`/`booting`/`restoring`/`gating`/`resuming` I am deliberately keeping the failure. Those computers *do* have a VM, or are mid-way through acquiring one, and it dies when this process exits — so the old `filter_map` on `handle` was not "correctly skipping" them, it was losing a guest silently because the handle had not landed on the computer yet. The aggregate error reads "some sandboxes could not be handed over and will be killed on exit", which is true for exactly those states, and the composer logs it rather than aborting on it. Under-reporting a lost guest seems the worse failure to me than a noisier rolling upgrade — but if the platform side would rather treat a slow boot as routine, that is a one-line change to the `Ok` arm and I am happy to make it.

computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
```diff
@@ -124,6 +124,29 @@ impl ComputerActor {
...
+                // stayed put, so the handover's own outcome is what decides.
+                Ok(()) => {
+                    // The data plane reads the agent straight off the snapshot
+                    // and never round-trips the mailbox, so the state alone
+                    // cannot stop it — and `detached` projects `Ready`, which
+                    // is exactly what `require_alive_agent` admits. Dropping
+                    // the agent is what keeps an exec or a file write out of a
+                    // guest the successor now owns; a stop and a release do
+                    // the same for the same reason.
+                    self.forget_agent();
```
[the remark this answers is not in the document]
PeronGH: Real, and partly fixed in 5b186b7c — thanks, this is the sharper version of the earlier agent finding.

Confirmed the mechanism: `agent_in` clones the whole snapshot and hands out the `Arc<dyn GuestAgent>` with no lock held across the RPC — deliberately, it is the data-plane seam — so nothing the actor does to the snapshot afterwards can invalidate a clone already in a caller's hand.

There are two windows, and they are not equally fixable:

**Closed.** `forget_agent` ran *after* `tasks.detach()`, which left the whole port call open — every reader arriving during the transfer took a working agent. Detach is the only flow whose ownership transfer happens inside an await, so this window was as wide as the driver's detach. The forget now runs first.

**Not closed.** A reader that cloned the `Arc` a moment before the forget still holds a working agent and can issue its RPC after ownership transferred. Closing that means invalidating an already-handed-out agent — either a revocation on `GuestAgent`, or routing the data plane through the mailbox, which is the hop this seam exists to avoid (an actor round-trip per exec is measured against the cold-start baselines). I did not want to reshape the data-plane seam inside a fix for the lifecycle race, so I have left it and named it in the code comment rather than silently narrowing the finding.

Worth noting the residual is bounded in a way the VM race was not: the straggler is a single in-flight exec or file write landing in a guest the successor owns and keeps running. The bug this PR fixes was a SIGKILL of that guest plus its TAP and CoW device being handed back underneath it.

The cost of reordering is that a refused handover must republish the agent — `usable` includes `dialable`, and nothing changed hands. Regressions: `the_agent_is_gone_before_the_handover_reaches_the_driver` (asserts the snapshot is already clear while the port call is parked mid-flight) and an added assertion in `a_refused_handover_leaves_the_computer_usable`.

computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
```diff
@@ -132,6 +132,45 @@ impl ComputerActor {
...
+                    | State::Stopped {}
+                    | State::Failed {}
+                    | State::Gone {}
+                    | State::Detached {} => Ok(()),
+                    // Mid-launch, mid-capture or mid-teardown — all of which
+                    // do have a VM, and all of which lose it when this process
+                    // exits. That is worth reporting rather than skipping: a
+                    // caller told `Ok` would believe a guest survived a
+                    // handover that never happened.
+                    _ => Err(self.wrong_state("Ready or Running")),
```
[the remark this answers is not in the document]
PeronGH: Fixed in c807a7f1 — and there is a second instance, which is why I did not take the per-state form.

`checkpointing` is not the only state in that arm whose projection lands inside the expected set. `State::Gating { claimed: true }` — a boot whose own `cmd` has taken the workload slot — projects `Running`, so it produced "expected Ready or Running, actual **running**" with the same contradiction. Giving `checkpointing` its own expected string, as suggested, would have left that one reading exactly as before.

So the expected side is now phrased as a property rather than a state list:

```rust
_ => Err(self.wrong_state("a computer with no launch, capture, or teardown in flight")),
```

That reads correctly for every state the arm refuses regardless of what it projects, needs no per-state arm, and matches how `refuse_handed_over` next to it already reads ("a computer this process still owns"). Deriving `actual` from the machine state instead was the other option you offered; I left it alone because `wrong_state` is shared by every refusal in the file and the machine's state names are not vocabulary the wire otherwise exposes.

One note on the regression, since it is the interesting part. It asserts the property — the expected side must not name the state being refused — rather than the wording, so rephrasing later will not break it. Its first form compared the two case-sensitively and **passed against the old message**: `SandboxState` renders lowercase, so `"Ready or Running".contains("ready")` is false. It now compares words case-insensitively, and I checked it fails against the string it replaces before restoring:

```
the refusal names the state it is refusing: expected Ready or Running, actual ready
```

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
```diff
@@ -543,25 +543,43 @@ impl SandboxManager {
...
+        // racing it is the same class of bug; `stop_sandbox` and
+        // `remove_sandbox` wait it out for exactly this reason. Its *failure*
+        // is not propagated the way theirs is: this is the process's last
+        // chance to save every guest it holds, and refusing the whole handover
+        // because the sweep failed would kill all of them to report a fault in
+        // something that has already finished.
+        if let Err(error) = self.await_reconcile().await {
+            warn!(%error, "the startup sweep failed; handing over anyway");
+        }
+        let computers: Vec<(SandboxId, Mailbox)> = self
```
AprilNEA: **[CONFIRMED · correctness] `detach_all` does not quiesce the manager — VMs still boot after it reports Ok**

Two gaps around this snapshot:

1. A computer whose `Detach` answers `Ok` *without* a transition (Provisioning, Paused, …) is not frozen. `create_sandbox` inserts the mailbox at `reserve_actor` and only sends `Command::Provision` after the awaited network setup, so a concurrent `detach_all` gets its `Ok` while the machine is in Provisioning (the door guard fires only on `State::Detached`), and the queued `Provision` then boots a full VM in an exiting process — killed by handle `Drop`, record left at `Starting`, reinstated as `Failed` by the successor, after the composer was told every guest was saved. Same shape for a `Resume` racing a Paused computer's `Ok`.
2. A sandbox registered *after* this read-lock snapshot is never sent `Detach` at all and dies unreported.

There is no shutdown latch on `SandboxManager`; the fix that closes both at once is a manager-level drain latch (set before the snapshot; refuses new registrations and lifecycle verbs), not more per-state arms.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
```diff
@@ -543,25 +543,43 @@ impl SandboxManager {
...
-            };
-            match detach.detach().await {
-                Ok(_) => info!(sandbox_id = %id, "handed the sandbox's vm to the next process"),
+        for (id, mailbox) in computers {
+            match mailbox.ask(&id, |reply| Command::Detach { reply }).await {
+                Ok(()) => info!(sandbox_id = %id, "handed the sandbox's vm to the next process"),
+                // The actor finished between the snapshot and the ask, so its
+                // computer is already gone — there is nothing left to hand
+                // over and nothing was lost.
+                Err(VmmError::NotFound(_)) => {}
```
AprilNEA: **[PLAUSIBLE · correctness] The `NotFound` swallow conflates "already removed" with "actor task died"**

`Mailbox::ask` maps both a closed mailbox *and* a dropped reply to `VmmError::NotFound` (`actor/mod.rs:107-110`). The actor is `tokio::spawn`ed unwatched with no `catch_unwind`, and its unregister/waiter-drain tail runs only on clean loop exit — so a panicked actor (e.g. via the `ComputerRuntime` mutex both the actor and flows `.lock().unwrap()`) leaves a map entry whose `ask` answers `NotFound`.

The comment here — "its computer is already gone … nothing was lost" — is false in exactly that case: the guest is live, never detached, SIGKILLed on process exit, and `detach_all` returns `Ok`. Distinguishing send-failure from reply-drop (or watching the actor's join handle) would keep the swallow honest.

computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
```diff
@@ -132,6 +148,56 @@ impl ComputerActor {
                     }
                 }
             }
+            Command::Detach { reply } => {
+                // Parked *before* the dispatch, like a removal and for a
+                // sharper reason: the handover is awaited inside the effect
+                // rather than spawned, so `Answer::Detached` always lands
+                // before this dispatch returns.
+                self.waiters.push((Answer::Detached, reply));
```
AprilNEA, line 181: **[CONFIRMED · correctness] Raw `waiters.push` bypasses `park()` — a successful handover can be reported as a failure**

`answer(Answer::Detached)` (commands.rs:342-355) unconditionally folds a stale `answer_error`/`unconfirmed` left by an earlier waiter-less flow into this reply. Reachable trigger: an idle-timer-driven pause (no command, no waiter) whose `Pausing` durable write is refused sets `answer_error` because `fail_every_waiter` answered 0 (effects.rs:326-327), and the Failure arm returns the machine to Ready. The next `Command::Detach` hands the VM over successfully, transitions to Detached — and replies `Err(Unavailable("...recording pausing failed..."))`.

`detach_all` then reports a lost guest that was in fact handed over, unretryably: the state is terminal, and a retry `Detach` now answers `Ok`, contradicting the first reply. Clearing the stale answer state when parking the Detach waiter (i.e. going through `park()`) closes it.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
```diff
@@ -543,25 +543,43 @@ impl SandboxManager {
...
     pub async fn detach_all(&self) -> Result<()> {
-        let live: Vec<(SandboxId, Arc<dyn VmHandle>)> = self
+        // The sweep adopts and kills VMs by the same deterministic names, so
+        // racing it is the same class of bug; `stop_sandbox` and
+        // `remove_sandbox` wait it out for exactly this reason. Its *failure*
+        // is not propagated the way theirs is: this is the process's last
+        // chance to save every guest it holds, and refusing the whole handover
+        // because the sweep failed would kill all of them to report a fault in
+        // something that has already finished.
+        if let Err(error) = self.await_reconcile().await {
```
AprilNEA: **[PLAUSIBLE · robustness] Unbounded `await_reconcile()` can starve the last-chance handover entirely**

`await_reconcile` is `rx.wait_for(Option::is_some)` with no deadline, and the sweep publishes only on completion; inside it, `handle.shutdown(Kill)` and the CoW teardown subprocesses (dmsetup/losetup) are unbounded (`reconcile.rs:744, 982, 1020` — only adoption has `ADOPT_TIMEOUT`). A restart-then-quick-stop during a botched redeploy parks `detach_all` here past the supervisor's kill deadline, and zero VMs are handed over — including fresh ones unrelated to the sweep.

A sweep *failure* is already downgraded to a warn; a bounded wait that falls through to the same warn-and-continue would fit the "last chance to save every guest" intent. (The pre-PR code went straight to the map.)

computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
```diff
@@ -124,6 +124,47 @@ impl ComputerActor {
...
+                let agent = self.snapshot_tx.borrow().agent.clone();
+                self.forget_agent();
+                match self.tasks.detach().await {
+                    // The machine has not moved yet: it emitted this effect
+                    // and stayed put, so the handover's outcome is what
+                    // decides.
+                    Ok(()) => self.queued.push_back(Event::Detached),
+                    // Stay where we are. The VM is still ours and still usable
+                    // — it simply dies with this process, which is what a
+                    // failed handover has always meant. Failing the computer
```
AprilNEA, line 166: **[CONFIRMED · robustness, latent] The only inline-awaited port call, with no timeout and no panic insulation**

`Effect::Detach` is awaited on the actor task itself, while every sibling flow runs under `tokio::spawn` with `JoinError` mapped to `VmmError::Process` (`inflight.rs:98-105`). Harmless for today's driver (`FcProcess::detach` is an atomic store + oneshot send — no await, no I/O, infallible), but a future driver whose detach does real work (fsync'd adoption record, remote control plane) turns each Detach into an unpreemptible actor stall — queued commands unserved, `detach_all`'s loop stuck — and a panic unwinds `run()`, skipping unregister and the waiter drain (which the `NotFound` swallow in `detach_all` then misreports as success).

The PR's "releasing ownership has no wait in it" invariant lives only in prose, two crates away from the async port it constrains; a bounded timeout here (or spawning like every other flow, with a no-persist failure arm) would encode it.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
```diff
@@ -543,25 +543,43 @@ impl SandboxManager {
...
             .collect();
 
         let mut failures = Vec::new();
-        for (id, handle) in live {
-            let Some(detach) = handle.detach() else {
-                continue;
-            };
-            match detach.detach().await {
-                Ok(_) => info!(sandbox_id = %id, "handed the sandbox's vm to the next process"),
+        for (id, mailbox) in computers {
```
AprilNEA: **[CONFIRMED · efficiency] Sequential asks: one wedged actor head-of-line-delays every remaining handover**

An actor inline-blocked in `Effect::AbortInflight` (up to ~20s: two 10s `HANDOFF_TIMEOUT` awaits, re-blocking per stall retry; emitted by Remove/TtlExpired/IdleExpired-Kill and the restore/gate failure unwinds) delays every computer behind it — precisely when the composer's drain deadline has already expired, which is the only time `detach_all` runs. A couple of TTL-expiry removals preempting wedged boots put 10–20s each in front of the loop, the kill timer fires mid-loop, and guests that were individually handoverable in microseconds die unasked.

`join_all`/`FuturesUnordered` over the `(id, mailbox)` pairs preserves per-computer serialization exactly (the actor is the serializer) and turns sum-of-stalls into max-of-stalls.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
```diff
@@ -543,25 +543,43 @@ impl SandboxManager {
...
+            match mailbox.ask(&id, |reply| Command::Detach { reply }).await {
+                Ok(()) => info!(sandbox_id = %id, "handed the sandbox's vm to the next process"),
```
AprilNEA: **[CONFIRMED · observability] This log fires for computers that had no VM to hand over**

Every `Ok` answer lands here — including Paused/Stopped/Failed/Provisioning/already-Detached computers, whose `Ok` means precisely that nothing changed hands (the `Detached` reply carries no payload to distinguish). A node with five paused and two stopped sandboxes logs seven handovers of VMs that did not exist; the pre-PR code logged only after a real `handle.detach()` succeeded.

The successor's on-call diagnosing a redeploy counts phantom guests and investigates a "lost handover" that never happened. A payload on the reply (`HandedOver | NothingToHandOver`) or logging from the effect keeps the forensic trail honest.

computer/arcbox-computer-runtime/src/lifecycle/machine.rs
```diff
@@ -337,9 +357,18 @@ impl ComputerLifecycle {
...
             }
+            // A handover is refused here, unlike in the rest of `active`. The
+            // capture runs in a sub-task that cloned the handle before it
+            // started, and `Detach` stands nothing down — the driver's detach
+            // only releases the reaper, leaving the handle fully functional —
+            // so the task would go on to freeze, snapshot and resume a guest
+            // this process no longer owns. Refusing costs the departing
+            // process's capture; accepting would corrupt the successor's
+            // guest, and the caller hears which computer it lost.
             Event::ClaimWorkload { .. }
```
AprilNEA, line 375: **[CONFIRMED · robustness, latent] `checkpointing` swallows `Event::Detach` but not `Event::Detached`**

`Event::Detached` bubbles to `active` and would transition a mid-capture machine straight to Detached. Unreachable today only because `Effect::Detach` is awaited inline and its `Detached` is drained within the same dispatch — but `harness::alphabet()` now includes `Detached`, so `explore()` generates Checkpointing→Detached nodes as *legitimate* rather than pinning the hole, and the day `Effect::Detach` becomes a spawned task (the shape of every sibling flow), a completion landing mid-capture detaches while the capture sub-task freezes/snapshots/resumes the successor's guest, with `detached` swallowing its `CaptureDone`.

One `Event::Detached` arm in this Handled list (or an invariant asserting Detached only fires from Ready/Running) closes it.

computer/arcbox-computer-runtime/src/lifecycle/machine.rs
```diff
@@ -337,9 +357,18 @@ impl ComputerLifecycle {
...
             | Event::Pause { .. }
             | Event::Checkpoint
+            | Event::Detach
```
AprilNEA, line 378: **[PLAUSIBLE · design] Refusing Detach mid-checkpoint forfeits a guest the actor could save by deferring**

The refusal is deliberate, documented, and test-pinned — this is a cost note, not a bug. But a user checkpoint 200ms from finishing when the drain expires costs the whole guest: Detach is refused `WrongState`, `detach_all` reports the computer lost, and the process exits and kills a guest that a deferred detach would have saved. The actor already owns the mechanism for exactly this shape: `pending_stop` parks a Stop behind an in-flight launch and serves it on completion. A `pending_detach` served on `CaptureDone` would absorb the sub-second wait — and Capturing even proceeds to Paused, where Detach answers `Ok`.

AprilNEA (commented):
Deep review (adversarially verified; 3 additional candidates were refuted during verification and dropped). 9 findings are inline on the diff; the 6 below anchor to lines outside the diff (or an unchanged file), so they live here.

---

**1. [CONFIRMED] `sandbox.rs:511` — port-forward RPCs are ungated after a handover** (unchanged file, same CORE-145 class this PR closes elsewhere)

`sandbox_network_identity` has no ownership/state gate and the lease survives Detach (nothing on the detach path clears `runtime.network`). A `SandboxPortForwardRequest` arriving after `detach_all` answered Ok (`guest/arcbox-agent/src/agent/linux/sandbox.rs:865-891, :930-951` — no mailbox round-trip) reads the lease off the snapshot and mutates PREROUTING/mangle for the *successor's* guest; the successor's startup cleanup later flushes the rules, so a port the API confirmed silently stops routing. Pre-existing before this PR, but the `Effect::Detach` comment half-acknowledges the class — the data plane's agent is dropped, this control path is neither mailbox-routed nor gated.

**2. [PLAUSIBLE] `flows/teardown.rs:285` — a Detach-less driver yields terminal Detached zombies**

A driver whose `VmHandle::detach()` is `None` (or a missing handle) makes `detach_vm` return `Ok`, so the machine still reaches terminal Detached and `detach_all` logs a successful handover. Both triggers are future-only today (the FC driver implements Detach; `detach_all` has no production caller yet), but the failure is ugly: live in-process VMs reported "handed to the next process", the composer exits without stopping them — and each computer is now locked in Detached (Stop/Remove(force)/SetLifecycle refused forever, listing as Ready), so a composer that aborts its shutdown after `detach_all` is left with unkillable, unremovable zombies for the life of the process. The `tasks.rs` doc's "there is no VM this process would have killed on the way out" is false for exactly the in-process class it names.

**3. [PLAUSIBLE] `machine.rs:325` — Detach from Running persists a free workload slot the old exec still occupies**

The record persists phase-only (`PersistPhase::Ready`), so the successor adopts an idle-Ready computer with the single-workload slot unclaimed (`machine.rs:92`), while the departing process's `run_session` (cloned agent Arc) keeps pumping stdin/output into the same guest until the old process exits. During a live-upgrade overlap — the exact scenario `detach_all` exists for — the successor's first exec starts a second workload in a guest still running the first. Pre-PR detach was equally ungated (narrowed inherited gap, not a regression); worth either refusing Detach from Running-with-claimed-slot or recording the claim in the record the successor reads.

**4. [CONFIRMED] `sandbox/lifecycle.rs:795` — Detached projects Ready, so `agent_in` answers 500 INTERNAL with a false message**

Post-detach, exec/file verbs pass the state gate (Ready) and the missing agent surfaces as `VmmError::Vsock` → 500 INTERNAL claiming the sandbox "has no running vm to reach" — while list/inspect report it Ready with an IP, mailbox verbs answer 412 `WrongState`, and the VM is in fact running, owned by the successor. Retry classifiers keyed on 412/423 treat it as a server fault. This is the concrete cost of patching the Detached-projects-Ready trade only at the actor's door: every snapshot-reading pre-flight must independently remember Ready-may-mean-detached. An internal `Detached` variant mapped to READY only at the API adapter would make every gate refuse naturally.

**5. [CONFIRMED] `actor/effects.rs:497` — timer cancellation on Detached is not sticky**

`arm()`/`rearm()` key the deadline timers off `state.to_public()`, and Detached projects Ready — so the `CancelTimer(Ttl)/CancelTimer(Idle)` pair the Detached transition emits is undone by any rearm (`timers_enabled.changed()`). Harmless today by two accidents (timers_enabled flips exactly once at startup; the `detached` leaf swallows TtlExpired/IdleExpired), but the `machine.rs` comment claims the cancellation is what protects the successor's guest, while the actual protection is the swallow. A Detached early-return in `arm()` (or in `deadlines::ttl_due/idle_due`) makes the cancellation authoritative.

**6. [CONFIRMED] `actor/mod.rs:237` — `ComputerSnapshot.handle` now has zero readers, and its doc invites reintroducing the bypass**

This PR deleted the last reader, yet the snapshot still publishes a fully functional handle under a Ready-projecting state, and the field's doc ("The running VM, for the graceful hand-over a process exit does") describes exactly the map-bypass CORE-145 removed. The snapshot is deliberately mailbox-free, so neither the door guard nor the detached state can stop a future reader that clones the handle and drives it. Dropping the field (or clearing it in the Detach effect alongside `forget_agent`, and rewriting the two doc sentences) closes the door instead of leaving it unlatched.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
PeronGH: Confirmed and fixed in 40cea200.

`Mailbox::ask` does map both failures onto `NotFound`, and `tokio::spawn(actor.run())` is unwatched with no `catch_unwind`, so the reported case is real: the entry stays, the guest stays alive, and the pass called it "already gone".

Distinguishing send-failure from reply-drop at the mailbox turned out to be the wrong seam — a reply dropped because a create's `ActorReservation` unwound before spawning an actor is a legitimate `NotFound`, so the two causes are not separable there. The registry answers it instead: the incarnation now travels with the mailbox, and `still_registered` (the inverse of `forget_computer`, beside it) decides. Every clean ending unregisters — the actor's loop tail, the record it forgets, the reservation's `Drop` — and a re-created id installs a fresh incarnation, so an incarnation *still present* whose mailbox no longer answers is exactly the actor that died without finishing. That one is now reported as an unhanded guest.

`registration_is_per_incarnation` pins the predicate, including the same-id-recreate case. The reported branch itself has no test: no public API can produce a panicked actor, and the poisoned-mutex point is why the predicate reads nothing the actor owns.

computer/arcbox-computer-runtime/src/lifecycle/actor/commands.rs
PeronGH: Confirmed and fixed in 2b47f7f3.

The trigger holds. `ready`'s pause persists `Pausing` with `Durability::Warn`, which softens an *unconfirmed* write and not a refused one, so a refused write reaches `fail_write`; an idle-driven pause has nothing parked, so `answer_error` is set and nothing later clears it.

One correction on the remedy: going through `park()` does not close it. `park` reports the stale error to the Detach caller *immediately* and never parks a waiter, so the handover then runs, succeeds, and its `Answer::Detached` finds nobody — the same false loss one step earlier. The stale state belongs to a flow that already ended with no listener, so it is now dropped where the Detach waiter parks, and logged, since nothing else can report it.

`a_handover_is_not_answered_by_an_earlier_flows_failure` reproduces your path end to end (a record tampered to `Starting` refuses `Pausing` and admits the `Ready` the failed pause reverts to, so the computer comes back usable with the failure still owed). It was checked red against the old code, failing with exactly `Unavailable("computer box recording pausing failed: ...")`.

`Remove`'s raw push carries the same hazard and predates this; left alone here, and noted in the comment — its answer at least stays consistent under a retry, which is the part that made this one unretryable.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
PeronGH: Confirmed and fixed in 40cea200.

Verified: only `adopt` is bounded (`ADOPT_TIMEOUT`); `handle.shutdown(Kill)` and `cow_manager.teardown_checked` have no deadline, and the wait itself is new in this PR — the pre-PR code went straight to the map.

The wait now runs under `SWEEP_WAIT_BUDGET` (10s) and falls through to the same warn-and-continue as the failure, which is the disposition you point at. The value mirrors the sweep's own per-computer adopt bound, on the reasoning that past it the sweep is not about to finish; the trade it makes — proceeding while the sweep is midway — is named at the constant, and is strictly better than losing every guest on the node to it.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
PeronGH: Confirmed and fixed in 40cea200.

`abort_inflight` is awaited inline on the actor task and can hold it for two `HANDOFF_TIMEOUT`s, so the head-of-line delay is real. One case worse than reported: `reserve_actor` installs the mailbox but does *not* spawn the actor, so an ask to a computer mid-create blocks on that create's entire network setup before anything reads it.

The asks now run concurrently over a `JoinSet`, which preserves per-computer serialization exactly as you say — the actor is the serializer — and turns sum-of-stalls into max-of-stalls.

No dedicated test: the fakes' detach is instant, so sum-vs-max is not observable through the harness, and the three pre-existing `detach_all` tests plus `a_handover_during_a_stop_is_refused_and_reported` are what pin the contract the change had to keep.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
PeronGH: Confirmed and fixed in 40cea200.

The `info!` moved to `detach_vm`, which is the only place that can tell the three cases apart: a real `detach.detach()` success logs at info, and "no handle" / "the driver holds its VMs in-process" log at debug. A reply payload would have been the alternative, but `Answer::Detached` travels through the typed `waiters` list, so carrying one would mean a second `capture_reply`-shaped special case for no forensic gain over logging at the source.

`detach_all` now logs a count of what it *asked*, phrased as such, rather than one line per `Ok`.

computer/arcbox-computer-runtime/src/lifecycle/actor/effects.rs
PeronGH: Confirmed and fixed in 2b47f7f3, in the half that can be fixed here.

Verified the premise: `FcHandle::detach` is `process.detach()` — an atomic store plus a oneshot send — so it is awaitless and infallible today, and the exposure is exactly the latent one you describe.

The port call is now bounded by `DETACH_TIMEOUT` (5s), which is the option that keeps the invariant inside this crate instead of leaving it as prose about a driver two crates away. A stall lands where a refusal already does: the computer stays ours, dialable and stoppable, and the actor keeps serving — `a_handover_that_does_not_return_fails_the_handover_alone` asserts all three.

Spawning it instead is still declined, for the reason the PR gives: the failure arm would need an intermediate state, and `failed` writes `Failed` over a record the successor's sweep has to read as `Ready`. That leaves the panic half uncovered — a panic inside the port call unwinds `run()` and skips the unregister — and it is now less dangerous than it was, because `detach_all` reads the leftover registration rather than trusting `NotFound` (see the thread on that swallow).

computer/arcbox-computer-runtime/src/lifecycle/machine.rs
PeronGH: Confirmed and fixed in 2b47f7f3.

Both halves of the report check out: `active` owns the `Detached` arm, `checkpointing` did not shadow it, and `a_handover_only_acts_where_a_live_handle_is_settled` pins `Detach` alone — so with `Detached` in the alphabet, `explore` was recording Checkpointing→Detached as a legitimate edge rather than flagging it.

`checkpointing` now swallows `Event::Detached` beside the `Detach` it refuses, with the reason at the arm, and `a_handover_completion_only_lands_where_the_handover_was_asked_for` walks every state: effects only from Ready/Running, and no other state moved by a stray completion. That test fails against the un-shadowed arm, so the hole is pinned rather than described.

computer/arcbox-computer-runtime/src/sandbox/lifecycle.rs
PeronGH: Confirmed, and deliberately not fixed here.

Both halves check out. `reserve_actor` installs the mailbox and the actor is spawned only after the awaited network setup, so a `Detach` can be answered `Ok` from `provisioning` and the queued `Provision` then boots a VM in an exiting process — and the door guard cannot catch it, since that path never enters `State::Detached`. The post-snapshot registration is true by construction, and the `Paused` + `Resume` variant is the same shape.

What is new in this PR is the false `Ok`: pre-PR the boot happened just the same, and a handle-less computer was skipped by the `filter_map` and reported as nothing at all. So the reported harm is real but it is a reporting regression on top of a pre-existing gap, and you have named the fix correctly — a manager-level drain latch, set before the snapshot, refusing new registrations and lifecycle verbs. That is new public surface on `SandboxManager` and a behavior the platform composer has to be able to rely on, so it wants its own change rather than riding a review pass on the transition.

Deferring it, tracked as a follow-up rather than fixed here. Leaving this thread open until that issue exists so it is not closed on a promise.

computer/arcbox-computer-runtime/src/lifecycle/machine.rs
PeronGH: Agreed as a cost, and deferring it.

The refusal is the conservative half of the trade: it costs the departing process's capture, where accepting corrupts the successor's guest. Your `pending_detach` shape is the right way to stop paying it — `pending_stop` is the precedent, and Capturing→Paused does answer `Ok`, so the deferral would land somewhere sane.

Not in this PR, though: it changes the behavior this PR just pinned with `a_handover_is_refused_while_a_capture_is_in_flight` and `a_handed_over_computer_refuses_every_teardown`, and a deferral needs its own budget decision (a capture 200ms from done is worth waiting for; one that has just started, with the drain deadline already expired, is not — and the actor has no deadline to weigh that against today).

Tracked as a follow-up rather than fixed here; leaving this thread open until that issue exists.

